Issue #5, opened 2020-03-17
Develop Login  Component


julianmrodri:
Develop the actual login functionality in the Marketplace, which allows to connect with your web3 browser wallet.

PR #6, opened 2020-03-17
[WIP] Web3 login, implements #5

vojtechsimetka:
This adds logging in with web3 wallets and locally running node.

TODO:
- [ ] Break the Account component, add tests
- [x] Make the Account component configurable (you can pass in your own list of providers that are accepted)
- [ ] Detect which web3 wallet there is and display correct name
- [ ] Polish the design - change the component look to be more RIF-like, shorten the account address
- [ ] Detect change of account event and act upon it


jurajpiar:
Solves #5 Develop Login Component

Issue #7, opened 2020-03-17
Create controller with test data


vojtechsimetka:
## Task
Create controller with test data. For now everything can be managed in a local storage. [Here](https://docs.google.com/spreadsheets/d/1tS4Ciz7KIQm5gySG2wvhRIabRAu46zK8bCl6pTa68tM/edit?usp=sharing) are some test data.

PR #15, opened 2020-03-24
Feature/cash controller #7 

jurajpiar:
Solving #7 Create controller with test data


src/api/rif-marketplace-cache/cacheController.test.ts
```diff
@@ -0,0 +1,62 @@
...
+        persistence.setItem(MarketListingType.domainListing, domainListing);
+    })
+
+    test('adds item to cache', async () => {
+        const newItem: DomainItemType = {
+            _id: 'test_id',
+            currency: 'TEST_COIN',
+            domain: 'test_domain',
+            price: 9999999,
+            price_usd: 99999,
```
vojtechsimetka: I would love to see this as `priceFiat` rather than USD. The currency should be configurable. That being said we should probably tackle that in a separate issue.

Issue #10, opened 2020-03-18
Browsing/Filtering: Hook up UI components to Controller


julianmrodri:
## Task

Hooking up the components with the controller. Controller and UI components need to be completed for this task.

Issue #8, opened 2020-03-18
Hook up cache to controller 


julianmrodri:
## Task
Hook up the cache to the controller (both controller and cache needs to be finished to be able to complete this task.

jurajpiar:
Blocked by [Write/test RNS integration #18](https://github.com/rsksmart/rif-marketplace-cache/issues/18)

PR #26, opened 2020-04-13
feat: connects filter, cache and adds rns buy forms #8 #10

jurajpiar:


jurajpiar:
Master-rebased.

julianmrodri:
Noticed when running that I only see one filter for Price, the one for example for Currency are not being displayed.

src/components/organisms/DomainsFilters.tsx
```diff
@@ -0,0 +1,100 @@
...
+                            }
+                        }
+                    })
+                }}
+            />
+            {/* <SelectFilter title='Currency' items={currencyItems}/> */}
+        </>
+    )
+}
+
```
julianmrodri, line 99: Would be nice to at least display the Currency filter (checkboxes) so we can see it in the UI.

src/components/molecules/table/SelectRowButton.tsx
```diff
@@ -0,0 +1,24 @@
+import React, { FC } from 'react'
```
farinalvaro, line 1: 4 spaces indentation

src/components/organisms/DomainsFilters.tsx
jurajpiar: done, but disconnected ie "dead"

Issue #27, opened 2020-04-13
Advanced Styling


julianmrodri:
We need the following general styling changes:
* Install required fonts
* Remove bootstrap from project
* Make site Responsive
* Spinner

PR #28, opened 2020-04-14
Feature/advanced styling

farinalvaro:
This Pull Request is related to the issue #27.
It styles the site globally and installs the `Rubik` font. It also removes unnecessary dependencies and files that we did not need to style the site.

## Known issues
- We can definitely improve those styles but I think it would be good to merge and clean other branches as well
- Make the site responsive for small devices
- Fix Modal styles


julianmrodri:
Looks pretty good to me. But we would need a more advanced review @jurajpiar  can you help so we merge this? Thanks!

jurajpiar:
on it.

PR #29, opened 2020-04-17
ci: added linter

vojtechsimetka:
Added CircleCI and linter rules according to what we agreed in our [naming discussion](https://hackmd.io/9rbfNqHwTpSw3ZZKJDXekQ). The linting rules follow the [tasegir linter rules](https://github.com/AuHau/tasegir/blob/master/config/eslintrc.js) because we can not overwrite them from the config and hence can't use the tasegir directly.


.eslintrc
```diff
@@ -1,54 +1,108 @@
...
+    }
+  },
+  'extends': ['standard', 'plugin:@typescript-eslint/recommended', 'airbnb'],
+  'env': {
+    'browser': true,
+    'commonjs': true,
+    'es6': true,
+    'jest': true,
+    'node': true,
+    'mocha': true
```
jurajpiar, line 16: don't think we're using mocha at all in this project.

julianmrodri:
This is great! Bunch of fixes that need to be done but really cool feature

vojtechsimetka:
Yes this test is flaky, PR #30 is removing the test as that part is fully rewritten.

jurajpiar:
As the CI check says: change the last commit type.

Issue #22, opened 2020-04-07
List Domain: Hookup UI Components to controller


julianmrodri:
Add required functionality to the Controller to support the List domain form components.

**User Story:** [List Domain for Sale](https://github.com/rsksmart/rif-marketplace-project/issues/11)

jurajpiar:
Not sure I understand this. Could you expand on this? Is it for the listing domains ie the "sell" section?

Issue #13, opened 2020-03-18
List Domain: Creating the forms to List domain for sale


julianmrodri:
Creating the forms for the List for Sale item and hooking it up with the controller (Needs as a dependency to have the controller finished).

**User Story:** [List Domain for Sale](https://github.com/rsksmart/rif-marketplace-project/issues/11)

PR #30, opened 2020-04-17
feat: adds the domain selling forms #13 #22

jurajpiar:
~~haven't done the set price, currency and usd price fields, yet.~~ Everything else should work....🤞


src/api/rif-marketplace-cache/cacheController.ts
```diff
@@ -1,26 +1,27 @@
...
-    const data = await services.offers.find({
+export const fetchMarketData = async (filters?: MarketFilter) => {
+    if (!service.current) throw Error('Not connected to a service');
+    return await (service.current as any).find({
         query: filters
     })
-    return maplistingToType(data.data, listingType);
-}
\ No newline at end of file
+}
```
farinalvaro: I really like this approach, but was wondering if it would be better to get the service object from a function that already checks if the service was created so we don't have to manually check it every time we want to use it. Something like `getService()` and if it's not created we can throw the error there, otherwise we return the service object.
What do you think?

jurajpiar: I just wanted to keep it simple, really.
I'm not sure what you mean, but I can only think of one use case when the appropriate service is already created and that is when you navigate to domains, then to landing and back to domains (or any other service, but only the same one as just previously visited). Otherwise it's always a different service.

jurajpiar: If I had one service creator then I'd have to pass that service around to each fetch function I'd talk to. This way (also I thought) I keep the respective domains (as in business domains not RNS domains.] ) separated.

src/components/molecules/JobDoneBox.tsx
```diff
@@ -0,0 +1,39 @@
+import React, { FC } from 'react';
```
farinalvaro, line 1: 4 spaces indentation on tsx file

src/components/molecules/JobDoneBox.tsx
```diff
@@ -0,0 +1,39 @@
...
+import { colors } from 'rifui/theme';
+
+export interface JobDoneBoxProps {
+    text: string,
+}
+
+const useStyles = makeStyles((theme: Theme) =>
+    createStyles({
+        body: {
+            border: '1px solid #008FF7',
```
farinalvaro, line 13: I think we should reference the `primary color` from the `rifui` theme here

jurajpiar: will do

src/components/pages/rns/buy/DomainPurchased.tsx
```diff
@@ -0,0 +1,48 @@
...
+import { useHistory } from 'react-router';
+import { ROUTES } from 'routes';
+
+const useStyles = makeStyles((theme: Theme) =>
+    createStyles({
+        actions: {
+            display: 'flex',
+            justifyContent: 'space-between',
+            flexDirection: 'row',
+            padding: '3em',
```
farinalvaro, line 15: I'd use `theme.spacing(X)` here

src/rifui/components/molecules/SwitchTabs.tsx
```diff
@@ -61,20 +62,15 @@ const useStyles = makeStyles((theme: Theme) => ({
   }
 }));
 
-const SwitchTabs: FC<SwitchTabsProps> = ({ label1, label2, onChange }) => {
+const SwitchTabs: FC<SwitchTabsProps> = ({ label1, label2, value: controlledValue, onChange }) => {
   const classes = useStyles();
-  const [value, setValue] = useState(0);
+  // const [value, setValue] = useState(controlledValue);
```
farinalvaro, line 67: The controlled value is really good!
Probably better to remove the old commented code

src/store/App/appReducer.ts
```diff
@@ -1,15 +1,23 @@
...
-  const newState = (!!userAction && userAction(state, payload)) || state
+  if (!!userAction) logger.debug('App action:', action)
+  const newState = (!!userAction && userAction(state, payload)) || state;
+
+  if (state !== newState) {
+    logger.debug('Prev state:', state)
+    logger.debug('Next state:', newState)
+  } else {
+    logger.debug('No change:', newState)
+  }
```
farinalvaro, line 20: Awesome!

src/routes.ts
```diff
@@ -3,10 +3,17 @@ export const ROUTES = {
   COMMUNICATIONS: '/communications',
   DATA_SERVICE: 'data_service',
   PAYMENTS: '/payments',
-  DOMAINS: '/domains',
+  DOMAINS: {
+    BUY: '/buy_domain',
+    SELL: '/sell_domain',
+  },
```
farinalvaro, line 9: It's probably better to have buy and sell under the same subpath, like `domains/buy` and `domains/sell` so we can easily know the active section of the site and for example highlight it in the header

src/components/pages/rns/index.ts
```diff
@@ -0,0 +1,12 @@
...
+import DomainOffersCheckoutPage from './buy/DomainOffersCheckoutPage';
+import DomainsCheckoutPage from './sell/DomainCheckoutPage';
+
+
+export {
+    DomainOffersPage as DomainsBuyPage,
+    MyDomainsPage as DomainsSellPage,
+    DomainOffersCheckoutPage,
+    DomainsCheckoutPage,
+}
```
farinalvaro, line 12: I'd only keep one of the namings to avoid confusion

jurajpiar: Leftover from renaming. Changed now.

src/routes.ts
jurajpiar: done

PR #32, opened 2020-04-20
Feature/advanced styling

farinalvaro:
Added responsive styles for tables, filters and header.
Fixed some broken styles like in the `connect wallet` modal

## Known issues
The homepage still needs to be styled


package.json
```diff
@@ -34,6 +34,7 @@
     "@feathersjs/socketio-client": "^4.5.2",
     "@material-ui/core": "^4.9.7",
     "@material-ui/icons": "^4.9.1",
+    "clsx": "^1.1.0",
```
jurajpiar, line 37: why do we need this library?

src/components/organisms/Header.tsx
```diff
@@ -18,33 +23,38 @@ const Login = () => (
...
     {
       title: 'Communications',
-      to: ROUTES.COMMUNICATIONS
+      to: ROUTES.COMMUNICATIONS,
+      icon: <ForumIcon />
     }
   ];
 
-  return <Navbar hreflogo={ROUTES.LANDING} items={navItems} login={Login} />;
+  return (<RUIHeader hreflogo={ROUTES.LANDING} items={headerItems} login={Login} />)
```
jurajpiar, line 57: Nicely done!

src/components/pages/rns/buy/DomainOffersCheckoutPage.tsx
```diff
@@ -149,7 +149,11 @@ const DomainOffersCheckoutPage = () => {
                     </CardActions>
                 }
             </Card>
-            {!!currentOrder && currentOrder.isProcessing && <TransactionInProgressPanel text='Listing the domain!' progMsg='The waiting period is required to securely buy your domain. Please do not close this tab until the process has finished.' />}
+            {!!currentOrder && currentOrder.isProcessing &&
+                <TransactionInProgressPanel
+                    text='Listing the domain!'
```
jurajpiar, line 154: My bad, this should read 'Buying the domain'

package.json
jurajpiar: I see you use it in button and header, but you can achieve the same thing with string interpolation. I'll leave my suggestion to those changes on the mentioned components.

src/rifui/components/organisms/Header.tsx
```diff
@@ -0,0 +1,217 @@
...
+        <div
+          role="presentation"
+          onKeyDown={toggleDrawer(false)}>
+          <AppBar
+            position="static"
+            className={
+              clsx(classes.mobileAppBar, {
+                [classes.mobileAppBarShift]: open,
+              })
+            }
```
jurajpiar, line 160: instead of using lib I suggest interpolation, such as:
```
`${classes.mobileAppBar} ${{ [classes.mobileAppBarShift]: open }}`
```

src/rifui/components/atoms/buttons/Button.tsx
```diff
@@ -24,8 +25,14 @@ const Button: FC<IButtonProps> = ({ className = '', block, shadow, rounded, chil
...
-    // TODO: consider using classnames library to conditionally render classes (https://github.com/JedWatson/classnames)
-    <MUIButton className={`${rounded ? classes.rounded : ''} ${block ? classes.block : ''} ${shadow ? '' : classes.noShadow} ${className}`}
+    <MUIButton
+      className={
+        clsx({
+          [classes.rounded]: rounded,
+          [classes.noShadow]: !shadow,
+          [classes.block]: block
+        },
+          className)}
```
jurajpiar, line 35: instead of using lib I suggest interpolation, such as:
```
`${classes.rounded} ${classes.noShadow} ${classes.block} ${className}`
```

src/rifui/components/molecules/Appbar.tsx
```diff
@@ -21,33 +21,34 @@ const useStyles = makeStyles((theme: Theme) => ({
...
   },
   navLinkContainer: {
-    minWidth: 50,
-    maxWidth: 200,
+    display: 'flex'
   },
 }));
 
-const Navbar: FC<NavbarProps> = ({ children, items, login, ...rest }) => {
+const AppBar: FC<AppBarProps> = ({ children, items, login, ...rest }) => {
```
jurajpiar, line 45: unused children

src/rifui/components/organisms/Header.tsx
```diff
@@ -0,0 +1,217 @@
...
+                      {navItem.title}
+                    </NavLink>
+                  </Typography>
+                ))
+              }
+            </div>
+            <div className={classes.loginContainer}>
+              <Login />
+            </div>
+          </Toolbar>
```
jurajpiar, line 146: You've got this in the Appbar component.

src/rifui/components/organisms/Header.tsx
```diff
@@ -0,0 +1,217 @@
...
+                    <NavLink to={headerItem.to} className={classes.mobileNavLink} activeClassName={classes.mobileNavLinkActive}>
+                      <ListItemIcon>
+                        {headerItem.icon}
+                      </ListItemIcon>
+                      <ListItemText primary={headerItem.title} />
+                    </NavLink>
+                  </ListItem>
+                ))}
+            </List>
+          </Drawer>
```
jurajpiar, line 209: Could be a good idea to extract this?

jurajpiar: I guess it's the whole <Hidden> block, though.

package.json
farinalvaro: You're right, good catch, thanks 🙂

src/rifui/components/organisms/Header.tsx
farinalvaro: After extracting the AppBar, this is the only thing that we have here and will be only used here so I think we can keep it here for now 🙂

PR #48, opened 2020-04-28
Rif ui extraction

farinalvaro:
This PR removes the dependency on the `rifui` folder and adds the dependency to the library that now lives in [RIF-UI](https://github.com/rsksmart/rif-ui) repository.
It also has the changes that the folder needed in order to be used as a library so we can review those changes here :).

## Alert
This PR has a dependency on the new library, pointing to the [feature/extraction](https://github.com/rsksmart/rif-ui/tree/feature/extraction) branch, please make sure it's still alive when reviewing or update the reference to master in the `package.json` file, just like this:

```
...,
"dependencies": {
...
    "@rsksmart/rif-ui": "github:rsksmart/rif-ui#feature/extraction",
...
},
```

## Known issues
- More like an issue to be solved in [RIF-UI](https://github.com/rsksmart/rif-ui) but we don't want to import files specifying the `dist` path.
- Similar as above, it would be ideal to import for example the theme from `@rsksmart/rif-ui/theme`.


jurajpiar:
Cannot successfully run it. The web3 account seems to be still undefined after login in.

src/rifui/components/molecules/ModalDialogue.tsx
```diff
@@ -21,13 +21,13 @@ const ModalDialogue: FC<ModalDialogueProps> = ({
...
     <Modal {...props}>
-      <>
+      <React.Fragment>
         <ModalHeader>
           <ModalTitle>{title}</ModalTitle>
         </ModalHeader>
         <ModalBody>{children}</ModalBody>
         <ModalFooter>{footer}</ModalFooter>
-      </>
+      </React.Fragment>
```
jurajpiar, line 30: Did you say you couldn't use <> for fragment for some reason?

farinalvaro: Yes, it would work here but as a library we can't compile it. Not sure about the reason but just for curiosity, this seems to be related: 
https://github.com/Microsoft/TypeScript/issues/20469
https://github.com/facebook/react/issues/12097

package.json
```diff
@@ -35,6 +35,7 @@
     "@material-ui/core": "^4.9.7",
     "@material-ui/icons": "^4.9.1",
     "@rsksmart/rif-marketplace-nfts": "github:vojtechsimetka/rif-marketplace-nfts",
+    "@rsksmart/rif-ui": "github:rsksmart/rif-ui#feature/extraction",
```
vojtechsimetka, line 38: I suggest we first merge the one on the `rif-ui` so that we take master branch.

src/components/atoms/Icon.tsx
```diff
@@ -1,19 +1,16 @@
 import React, { FC } from 'react';
-import rifDir from 'rifui/assets/images/rifDir.png';
-import rifCom from 'rifui/assets/images/rifCom.png';
-import rifGat from 'rifui/assets/images/rifGat.png';
-import rifMar from 'rifui/assets/images/rifMar.png';
-import rifPay from 'rifui/assets/images/rifPay.png';
-import rifSto from 'rifui/assets/images/rifSto.png';
+import {
```
vojtechsimetka, line 2: why is this not part of the library?

src/components/templates/MarketPageTemplate.tsx
```diff
@@ -2,9 +2,9 @@ import React, { FC, useContext } from 'react';
 import MarketFilter from 'components/templates/marketplace/MarketFilter';
 import Marketplace, { TableHeaders } from 'components/templates/marketplace/Marketplace';
 import { MarketItemType } from 'models/Market';
-import { Grid } from 'rifui';
 import { makeStyles, Theme } from '@material-ui/core/styles';
-import { Web3Store } from 'rifui/providers/Web3Provider';
+import { Grid } from '@rsksmart/rif-ui';
```
vojtechsimetka, line 6: Minor issue that will be solved with linter (I would not worry about it now), but you want to usually to group the import from the same library into one. So `{ Grid, Web3Store ...}`

farinalvaro: Totally. The point is that I would like to import `Web3Store` from `@rsksmart/rif-ui/providers` but first we need to solve that in the UI library, but for now I'll gorup this.

src/components/atoms/Icon.tsx
farinalvaro: Agree. Will add this to the library in the next PR.

PR #62, opened 2020-05-04
feat: adds price bounds fetch, changing filter #50

### LINKED ISSUE #50: Price filter not setting upper bound
The upper bound of the filter should be set to the Max price available in the listings. Now its fixed in 100 RIF.

jurajpiar:
Closes #50 
Adds a request to the cache to get the min and max price. Calling it from componentWillMount of the DomainOfferFilters, it sets edge values of the RangeFilter component.
As mentioned in the issue this closes, a [rif-ui issue](https://github.com/rsksmart/rif-ui/issues/10) prevents us from changing the inputs associated with the bounds of the range slider programmatically.
The result is that, though the slider itself updates its values, as well as does the marketplace (showing items according to min max boundaries), the <input> parts of the slider are not representing this change.
As soon as this is fixed in the rif-ui, the behaviour should be as expected.

### DIFF
--- a/src/api/rif-marketplace-cache/domainsController.ts
+++ b/src/api/rif-marketplace-cache/domainsController.ts
@@ -26,7 +26,6 @@ export interface DomainTransferItem {
     tokenId: string,
 }
 
-// TODO: refactor DomainOffer, Domain dates
 const mappings = {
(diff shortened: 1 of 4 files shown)


julianmrodri:
LGTM - Tested locally

PR #71, opened 2020-05-07
feat: adds rsktestnet config

jurajpiar:


src/ui-config.json
```diff
@@ -4,5 +4,11 @@
         "rnsDotRskOwner": "0x4bf749ec68270027C5910220CEAB30Cc284c7BA2",
         "rnsReverseRegistrar": "0xe982E462b094850F12AF94d21D470e21bE9D0E9C",
```
julianmrodri, line 5: Shouldnt we fix this other configuration? To avoid confusion?

julianmrodri: They should both have the same varilables (I guess the one for Testnet are the correct)

src/ui-config.json
```diff
@@ -4,5 +4,11 @@
...
         "marketplace": "0x22d5C8BdD4346b390014a07109a8F830094d4abf"
-    }
-}
\ No newline at end of file
+    },
+    "rsktestnet": {
+        "rif": "0x19f64674D8a5b4e652319F5e239EFd3bc969a1FE",
+        "rnsDotRskOwner": "0xca0a477e19bac7e0e172ccfd2e3c28a7200bdb71",
+        "rnsNameResolver": "0x8587385ad60038bB181aFfDF687c4D1B80C4787e",
```
vojtechsimetka, line 11:
```suggestion
        "rnsReverseRegistrar": "0x8587385ad60038bB181aFfDF687c4D1B80C4787e",
```

src/ui-config.json
vojtechsimetka: Good eye

vojtechsimetka: Even though the correct name would actually be `rnsNameResolver`

julianmrodri: I thought rnsNameResolver was the one being used

PR #72, opened 2020-05-07
Adding installation instructions

julianmrodri:
Adding installation instructions and explanation of Config files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,11 +44,11 @@ Setup config file `ui-config.json` which contains the required Contract addresse
 Example:
 
 ```
-"regtest": {
-    "rif": "0x67B5656d60a809915323Bf2C40A8bEF15A152e3e",
-    "rnsDotRskOwner": "0x4bf749ec68270027C5910220CEAB30Cc284c7BA2",
+"rskTestnet": {
+    "rif": "0x19f64674D8a5b4e652319F5e239EFd3bc969a1FE",
+    "rnsDotRskOwner": "0xca0a477e19bac7e0e172ccfd2e3c28a7200bdb71",
     "rnsNameResolver": "0x9561C133DD8580860B6b7E504bC5Aa500f0f06a7",
-    "marketplace": "0x22d5C8BdD4346b390014a07109a8F830094d4abf"
+    "marketplace": "0x8587385ad60038bB181aFfDF687c4D1B80C4787e"
   }
 ```
 
--- a/README.md
+++ b/README.md
@@ -5,11 +5,63 @@
 \*\*Warning: This project is in alpha state.
 
 ## Install
-For environment setup please see [Rif Marketplace Dev](https://github.com/rsksmart/rif-marketplace-dev#local-development-environment-for-rif-marketplace-services) project.
 
-## Configuration
+Download and setup the RIF Marketplace UI
+```
+git clone git@github.com:rsksmart/rif-marketplace-ui.git
+
+cd rif-marketplace-ui
+```
+
+Install the dependencies
+
+```
+npm i
+```
+
+Setup `.env` file with the following configuration:
+
+ - `REACT_APP_CACHE_ADDR` : Cache connection URI
+ - `REACT_APP_LOG_LEVEL`: React logging level
+ - `REACT_APP_NETWORK`: Network Identified (should be defined in `ui-config.json`) 
+ 
+Example:
+
+```
+REACT_APP_CACHE_ADDR=http://cacheserver:3030   
+REACT_APP_LOG_LEVEL=error
+REACT_APP_NETWORK=rskTestnet
+```
+
+Setup config file `ui-config.json` which contains the required Contract addresses for all networks. 
+
+- RNS related:
+    -  `rif`: address of the deployed RIF token contract
+    - `rnsDotRskOwner` : address of the deployed RNS Owner contract 
+    - `rnsNameResolver`: address of the deployed RNS Name resolver contract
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -5,12 +5,64 @@
...
+
+```
+npm i
+```
+
+Setup `.env` file with the following configuration:
+
+ - `REACT_APP_CACHE_ADDR` : Cache connection URI
+ - `REACT_APP_LOG_LEVEL`: React logging level
+ - `REACT_APP_NETWORK`: Network Identified (should be defined in `ui-config.json`)
```
vojtechsimetka, line 26:
```suggestion
 - `REACT_APP_NETWORK`: Network Identified. Following options are already defined: `ganache` and `rsktestnet`.(This network needs to be defined in `src/ui-config.json`) 
```

README.md
```diff
@@ -5,12 +5,64 @@
...
+ 
+Example:
+
+```
+REACT_APP_CACHE_ADDR=http://cacheserver:3030   
+REACT_APP_LOG_LEVEL=error
+REACT_APP_NETWORK=rskTestnet
+```
+
+Setup config file `ui-config.json` which contains the required Contract addresses for all networks.
```
vojtechsimetka, line 36:
```suggestion
If you want your own configuration, please add it to the config file `src/ui-config.json`. 
```

jurajpiar, line 36: Maybe should say something like: "This file is generated by the environment setup (Rif Marketplace Dev) as described above" (and would put the env setup as the first thing as ppl are likely to start putting all this in before reading that line if it is at the bottom)

README.md
```diff
@@ -5,12 +5,64 @@
...
+
+Example:
+
+```
+"rskTestnet": {
+    "rif": "0x19f64674D8a5b4e652319F5e239EFd3bc969a1FE",
+    "rnsDotRskOwner": "0xca0a477e19bac7e0e172ccfd2e3c28a7200bdb71",
+    "rnsNameResolver": "0x8587385ad60038bB181aFfDF687c4D1B80C4787e",
+    "marketplace": "0x712025c9fe1ca55296ce439c4ac04019c266a5b1"
+  }
```
vojtechsimetka, line 52:
```suggestion
"custom": {
    "rif": "0x...",
    "rnsDotRskOwner": "0x...",
    "rnsNameResolver": "0x...",
    "marketplace": "0x..."
  }
```

README.md
```diff
@@ -5,12 +5,64 @@
...
+Setup `.env` file with the following configuration:
+
+ - `REACT_APP_CACHE_ADDR` : Cache connection URI
+ - `REACT_APP_LOG_LEVEL`: React logging level
+ - `REACT_APP_NETWORK`: Network Identified (should be defined in `ui-config.json`) 
+ 
+Example:
+
+```
+REACT_APP_CACHE_ADDR=http://cacheserver:3030
```
jurajpiar, line 31: Wouldn't confuse users with http://cacheserver:3030. Just put http://localhost:3030 as you mention this addr bellow anyway.

README.md
julianmrodri: But the file does not to be generated by the Dev environment. If you just want to deploy it and run it against Mainnet you dont need the Dev environment right? this is to run stand alone

README.md
julianmrodri: I does not need to be Localhost. this is not meant to be ran with Cache in same server. Maybe something more clear like  CACHE_SERVER_URL oir somethign like that?

README.md
vojtechsimetka: @jurajpiar this file is versioned controlled and should not be generated in crude version like it is done until now. It will contain different configurations for different networks. This way you can easily switch between them. Later on, the UI will actually automatically detect which network it is connected to based on the networkId or chainId. See https://github.com/rsksmart/rif-ui/blob/master/src/providers/Web3Provider.tsx#L39-L64

vojtechsimetka:
@julianmrodri please use the right commits. We can squash this but in general we will require them (and the CI will enforce them). See https://www.conventionalcommits.org/en/v1.0.0/

README.md
vojtechsimetka: But it is understandable for anyone that it can be changed. `cacheserver` is just wrong and magical constant

julianmrodri: ok moving back to Localhost

README.md
```diff
@@ -5,12 +5,63 @@
...
+
+```
+npm i
+```
+
+Setup `.env` file with the following configuration:
+
+ - `REACT_APP_CACHE_ADDR` : Cache connection URI
+ - `REACT_APP_LOG_LEVEL`: React logging level
+ - `REACT_APP_NETWORK`: Select the Network to connect (should be defined in `ui-config.json`)
```
vojtechsimetka, line 26:
```suggestion
 - `REACT_APP_NETWORK`: Select the Network to connect (must be defined in `ui-config.json`) 
```

src/ui-config.json
```diff
@@ -2,7 +2,7 @@
     "ganache": {
         "rif": "0x67B5656d60a809915323Bf2C40A8bEF15A152e3e",
         "rnsDotRskOwner": "0x4bf749ec68270027C5910220CEAB30Cc284c7BA2",
-        "rnsReverseRegistrar": "0xe982E462b094850F12AF94d21D470e21bE9D0E9C",
+        "rnsNameResolver": "0x9561C133DD8580860B6b7E504bC5Aa500f0f06a7",
```
vojtechsimetka, line 5:
```suggestion
```

Issue #108, opened 2020-05-12
Move FAQ Expansion Pannels to rif-ui library


farinalvaro:
The components used to display the FAQ content should be moved to `rif-ui` and consumed here from there.

Issue #80, opened 2020-05-10
FAQ Section


julianmrodri:
Create a FAQ section based on the proposed designed. This section will be accessed from the Footer.

PR #109, opened 2020-05-12
feat: created FAQ page

farinalvaro:
This PR solves #80 - Created FAQ page

## Known issues
Some components and functionalities included here should go on `rif-ui` as mentioned in #108
The point is that we would need to update `rif-ui` components that are being used in other places (i.e. the filters)


farinalvaro:
Converted to draft until we have the right content, so we don't have to create another PR to update the content.

PR #125, opened 2020-05-18
feat: Canceling Domain functionality

julianmrodri:
Main goal: Adds the Canceling a Domain functionality.

Changes:  
- Added IconButton X for Canceling a Listed Domain.
- Adapted routes and flow to be able to Cancel Domain (requires two transactions - The "unapproval" of the Token and the actual "unplace" Transaction)
- Two New pages:
   - rns/cancel/DomainCanceled.tsx
   - rns/cancel/CancelDomainCheckoutPage.tsx

### DIFF
--- a/src/components/Routes.tsx
+++ b/src/components/Routes.tsx
@@ -15,9 +15,9 @@ import {
   DomainOffersCheckoutPage,
   DomainListed,
   CancelDomainCheckoutPage,
-  DomainCanceled
+  DomainCanceled,
+  SoldDomainsPage
 } from './pages/rns';
-import SoldDomainsPage from './pages/rns/sell/SoldDomainsPage';
 
 
 const logger = Logger.getInstance();
--- a/src/components/Routes.tsx
+++ b/src/components/Routes.tsx
@@ -8,11 +8,17 @@ import { FAQPage } from './pages/faq';
 import { StoragePage } from './pages/storage';
 import AboutPage from './pages/AboutPage';
 import {
-  MyDomainsPage, DomainListed,
-  DomainsCheckoutPage, DomainOffersCheckoutPage,
-  SoldDomainsPage, DomainOffersPage,
-  DomainPurchased
-} from './pages'
+  DomainsCheckoutPage,
+  MyDomainsPage,
+  DomainOffersPage,
+  DomainPurchased,
+  DomainOffersCheckoutPage,
+  DomainListed,
+  CancelDomainCheckoutPage,
+  DomainCanceled
+} from './pages/rns';
+import SoldDomainsPage from './pages/rns/sell/SoldDomainsPage';
+
 
 const logger = Logger.getInstance();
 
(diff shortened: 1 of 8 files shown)


jurajpiar:
Works well, just needs some UI touches. Also would be good to use existing components, but if needs closing quickly then that can be done in the refactor.

src/components/pages/rns/sell/MyDomainsPage.tsx
```diff
@@ -5,13 +5,14 @@ import MarketPageTemplate from 'components/templates/MarketPageTemplate';
...
 import { useHistory } from 'react-router';
-import { Web3Store } from '@rsksmart/rif-ui';
+import { Web3Store, IconButton } from '@rsksmart/rif-ui';
 import { ROUTES } from 'routes';
 import { MARKET_ACTIONS } from 'store/Market/marketActions';
 import MarketStore, { TxType } from 'store/Market/MarketStore';
 import { Domain } from 'models/marketItems/DomainItem';
 import AddressItem from 'components/molecules/AddressItem';
 import { CombinedPriceCell } from 'components/molecules';
+import ClearIcon from '@material-ui/icons/Clear';
```
jurajpiar, line 15: Shouldn't really use material ui directly. Instead this should be exported by rif-ui.

julianmrodri: Yes good point. Ill work with @farinalvaro  on getting that Icon to the RIF-UI lib.

vojtechsimetka: I actually disagree. Because we don't want to forward all the things like `Grid, Icons` etc from rif-ui. It is unnecessary to do this wrapping. We would have to potentially update rif-ui each time there is new material-ui release.

I suggest we go the other way. `rif-ui` enhances material ui and adds only styling through theme and component material ui is missing.

julianmrodri: Thanks. Ill leave it as is then.

src/components/Routes.tsx
```diff
@@ -8,11 +8,17 @@ import { FAQPage } from './pages/faq';
...
-} from './pages'
+  DomainsCheckoutPage,
+  MyDomainsPage,
+  DomainOffersPage,
+  DomainPurchased,
+  DomainOffersCheckoutPage,
+  DomainListed,
+  CancelDomainCheckoutPage,
+  DomainCanceled
+} from './pages/rns';
```
farinalvaro, line 19: I would keep importing from `pages` so we don't have to worry if we move the files in the subdirectories

src/components/pages/rns/cancel/CancelDomainCheckoutPage.tsx
```diff
@@ -0,0 +1,206 @@
...
+const marketPlaceAddress = contractAdds[NETWORK].marketplace.toLowerCase();
+
+const useStyles = makeStyles((theme: Theme) =>
+    createStyles({
+        card: {
+            width: 491,
+            height: "fit-content",
+            padding: 80,
+            paddingTop: 44,
+            paddingBottom: 69,
```
farinalvaro, line 32: I would suggest to use theme.spacing(x), where spacing(1) = 8px by default

jurajpiar:
I wonder whether from UX perspective it does make much sense to go to a different page (DomainCancelled) after the cancel transaction. Surely, the user would rather go back to the list than to a confirmation page?
Having said that, that could potentially apply to all transactions, but that would probably drastically change the design, like using modals instead of pages for transactions and confirmation.

src/components/pages/rns/cancel/CancelDomainCheckoutPage.tsx
jurajpiar: Done by ito in f94e223.

PR #144, opened 2020-05-21
feat: removes truffle contract

jurajpiar:
This also fixes the re-surfaced issue with FireFox


src/components/pages/rns/cancel/CancelDomainCheckoutPage.tsx
```diff
@@ -85,7 +83,7 @@ const CancelDomainCheckoutPage = () => {
       web3,
     },
   } = useContext(Web3Store)
-  const Contract = (contract) => ContractWrapper(contract, (web3 as Web3), (account as string))
+  // const Contract = (contract) => ContractWrapper(contract, (web3 as Web3), (account as string))
```
farinalvaro, line 86: We could remove this line

julianmrodri: Yes lets remove the line I agree.

julianmrodri, line 86:
```suggestion
```

PR #155, opened 2020-05-27
feat: adds support for events coming from cache

### LINKED ISSUE #51: Capture events from the cache service
https://docs.feathersjs.com/api/client/socketio.html#listening-to-events

jurajpiar:
This PR adds a package @material-ui/lab to allow us to use their Alert and later Pagination components.
This means that you must `npm i` after checking out this PR.
Closes #51 
To test this you will need [Cache PR#130](https://github.com/rsksmart/rif-marketplace-cache/pull/130)

### DIFF
--- a/src/components/molecules/InfoBar.tsx
+++ b/src/components/molecules/InfoBar.tsx
@@ -1,0 +1,48 @@
+import { Button, ButtonProps, createStyles, makeStyles, Snackbar, Theme } from '@material-ui/core'
+import MuiAlert, { Color as Severity } from '@material-ui/lab/Alert'
+import React, { FC } from 'react'
+
+export interface InfoBarProps {
+    isVisible: boolean
+    text: string
+    button: ButtonProps
+    buttonText: string
+    type: Severity
+}
+
+const useStyles = makeStyles((theme: Theme) =>
+    createStyles({
+        alert: {
+            alignItems: 'center',
+            display: 'flex',
+            flexDirection: 'row',
+        },
+        button: {
+            padding: 0
+        }
+    }),
+)
+
+const InfoBar: FC<InfoBarProps> = ({ isVisible, text, button, buttonText, type }) => {
+    const classes = useStyles();
+    const action = <Button {...button} />
+    return (
+        <Snackbar
+            anchorOrigin={{ vertical: 'top', horizontal: 'center' }}
+            open={isVisible}
+        >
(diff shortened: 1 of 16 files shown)


src/components/molecules/InfoBar.tsx
```diff
@@ -0,0 +1,44 @@
+import {
+  Button, ButtonProps, createStyles, makeStyles, Snackbar, Theme,
+} from '@material-ui/core'
+import MuiAlert, { Color as Severity } from '@material-ui/lab/Alert'
+import React, { FC } from 'react'
+
+export interface InfoBarProps {
+  isVisible: boolean
+  text: string
+  button: ButtonProps
```
farinalvaro, line 10: May we rename this prop to `buttonProps` to avoid confusion?

src/components/molecules/InfoBar.tsx
```diff
@@ -0,0 +1,44 @@
+import {
+  Button, ButtonProps, createStyles, makeStyles, Snackbar, Theme,
```
farinalvaro, line 2: I think we could import `Button` and `ButtonProps` from rif-ui. It would make no difference right now but we could get benefited from future changes to the library. What do you think?

src/store/Market/marketReducer.ts
```diff
@@ -116,6 +121,21 @@ const marketActions: MarketActionsType = {
...
+      ...state,
+      metadata: {
+        ...metadata,
+        [listingType]: {
+          ...metadata[listingType],
+          updatedTokens: [...metadata[listingType].updatedTokens, ...[payload.updatedTokenId]],
+        },
+      },
+    } : state
+  },
```
farinalvaro, line 138: wow, this looks great!

jurajpiar:
Waiting for Cache to be finished

src/components/molecules/InfoBar.tsx
jurajpiar: that was deliberately not buttonProps, but I'm open to it. My reasoning was that buttonProps is kinda an ugly and the "props" in it is somewhat extraneous since its type tells you that it is indeed a ButtonProps type. If fact I'd like us to propagate this sort of naming across the project.
I'm happy to discuss it tho.

src/components/molecules/InfoBar.tsx
jurajpiar: Yeap that sounds good.

PR #161, opened 2020-06-01
feat: adding Tesnet env file

julianmrodri:
Adding Env file for Testnet environment. Actual value for Cache URL has to be confirmed by Devops.


julianmrodri:
Closing as decided not to include this feature and simply leave one .env file.

PR #193, opened 2020-06-17
feat: add link to RNS Manager

julianmrodri:
* Adds link to RNS Manager after you Buy a domain.
* RNS Manager URL has to be defined in the `ui-config.json` file.
* Checks that the URL is valid otherwise disables the button.
* Postpones Cleanup of `currentOrder` to the Transaction confirmation page.

Note: To test this PR you need to:
1) Run the RNS Manager locally with the LATEST `develop` branch.
2) If you get an error when doing `npm start` in the RNS manager run: `npm i webpack@4.28.3` in the Manager project this should fix the issue. Then try again.
3) Setup the correct RNS manager URL in the `ui-config.json` file of the MArketplace UI project. (samples provided in the default file).

4) **BUY** a domain and then click on the `Admin my domain` button of the confirmation page. You should get a new tab opened directly on the manager page (if you didnt switch the account in between, you have to be with the samne account that just bought the domain)

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,7 +51,7 @@
   },
   "devDependencies": {
     "@testing-library/jest-dom": "^5.9.0",
-    "@testing-library/react": "^10.2.1",
+    "@testing-library/react": "^10.3.0",
     "@testing-library/user-event": "^11.3.1",
     "@typechain/web3-v1": "^1.0.0",
     "@types/jest": "25.2.3",
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
     "web3-utils": "^1.2.8"
   },
   "devDependencies": {
-    "@testing-library/jest-dom": "^5.9.0",
+    "@testing-library/jest-dom": "^5.10.1",
     "@testing-library/react": "^10.3.0",
     "@testing-library/user-event": "^11.3.1",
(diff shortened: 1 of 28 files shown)


jurajpiar:
I'm unable to place a domain now:

Screenshot 2020-06-19 at 11 41 21

package.json
```diff
@@ -45,6 +45,7 @@
     "react-scripts": "3.4.1",
     "socket.io-client": "^2.3.0",
     "typescript": "^3.9.5",
+    "valid-url": "^1.0.9",
```
jurajpiar, line 48: I don't think we should introduce yet another package just to check URL. It seems it doesn't add that much value to the project. We could just check whether it exists or not. the rest is a responsibility of those implementing the marketplace project.

src/components/pages/rns/buy/DomainPurchased.tsx
```diff
@@ -42,6 +43,7 @@ const DomainPurchased: FC<{}> = () => {
           color="primary"
           variant="contained"
           rounded
+          disabled={!validUrl.isUri(rnsManagerUrl)}
```
jurajpiar, line 46: Should we still want to use this package tho, we should use the correct method **isHttpUri**, or better still **isWebUri**.

PR #210, opened 2020-06-30
refactor: normalises stores

### LINKED ISSUE #152: MarketStore refactor
Extract domain specific store from MarketStore but keep it for common needs (store, actions, reducers)

jurajpiar:
Closes [#RMKT-183](https://rsklabs.atlassian.net/jira/software/projects/RMKT/boards/60?selectedIssue=RMKT-183)
Closes #152 

The idea here is to let each store handle the communication with the services. This way the components are cleaner with less side effects, ideally just presenting the data stored in the stores. This should also make it easier and cleaner to do error handling and any other system-to-user communication.

Known insufficiencies:
- all store providers are mounted at the start of the application. this should change to have been only mounted when needed
- sold filters not actually connected (this should be a different issue tho, where the own, placed and sold domains get bumped up a level away from the filters, not dissimilar to the tx types)


src/components/organisms/rns/sell/SoldDomains.tsx
```diff
@@ -111,20 +66,14 @@ const SoldDomains: FC<{}> = () => {
...
-    currency: 'Currency',
-    sellingPrice: 'Selling price',
-    soldDate: 'Selling date',
-  }
-
   return (
     <MarketPageTemplate
       filterItems={<DomainFilters />}
-      itemCollection={collection}
+      itemCollection={collection as any}
```
farinalvaro, line 72: Why do we need this `as any`? Do you think it could be a good idea to split the MarketPageTemplate in the future and have abstract content and filters? Both of them being React components/elements

src/components/pages/rns/buy/DomainOffersCheckoutPage.tsx
```diff
@@ -119,16 +122,19 @@ const DomainOffersCheckoutPage: FC<{}> = () => {
...
+      dispatch({
+        type: 'CLEAR_ORDER',
+      } as any)
     }
-  }, [currentOrder, isPendingConfirm, history])
+  }, [order, isPendingConfirm, history, dispatch])
 
-  if (!currentOrder) return null
+  if (!order) {
+    history.replace(ROUTES.LANDING)
```
farinalvaro, line 135: Could we just return `history.replace(ROUTES.LANDING)`?

src/components/pages/rns/buy/DomainOffersPage.tsx
```diff
@@ -127,14 +98,16 @@ const DomainOffersPage: FC = () => {
       }
 
       return displayItem
-    })
+    }) as any // TODO: remove as any
```
farinalvaro, line 101: I know we had a reason for this but I can't remember why we needed this as any. Do you remember?

src/store/Market/rns/OffersStore.tsx
```diff
@@ -0,0 +1,151 @@
...
+        payload: { refresh: true },
+      } as any)
+      setIsReady(true)
+    }
+  }, [api, isReady, needsRefresh])
+
+  useEffect(() => {
+    const { fetchPriceLimits } = api
+
+    if (isReady && needsRefresh && !isLimitsSet) {
```
farinalvaro, line 104: Could you please add some comments about what the flags mean? Also noticed that we use the same flags on the others stores, may we extract this functionallity in the future?

src/utils/typeUtils.ts
```diff
@@ -0,0 +1 @@
+export type Modify<T, R> = Omit<T, keyof R> & R
```
farinalvaro, line 1: Nice!

src/store/storeUtils/reducer.ts
```diff
@@ -0,0 +1,31 @@
+import Logger from 'utils/Logger'
+import {
+  StoreDispatcher, StorePayload, StoreState, StoreActions, StoreAction, StoreReducer,
+} from './interfaces'
+
+const logger = Logger.getInstance()
+
+const storeReducerFactory = (initialState: StoreState, actions: StoreActions, errorHandle?: Function): StoreReducer => (state = initialState, dispatcher: StoreDispatcher<StorePayload>) => {
```
farinalvaro, line 8: Love this extraction for reusability!

src/api/models/apiController.ts
```diff
@@ -0,0 +1,40 @@
...
+}
+
+export abstract class AbstractAPIController implements Omit<APIController, 'fetch'> {
+  path!: string
+
+  service!: Service<any>
+
+  connect = (clientOverride?: Application<any>) => {
+    const app = clientOverride || client
+    try {
```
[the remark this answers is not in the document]
jurajpiar: yes, I'll remove the try/catch. thanks for reminder.

src/api/rif-marketplace-cache/rns/domains.ts
```diff
@@ -0,0 +1,51 @@
...
+export class DomainsController extends AbstractAPIController implements RnsAPIController {
+  path = domainsAddress
+
+  fetch = async (filters: Partial<RnsFilter>): Promise<RnsDomain[]> => {
+    if (!this.service) throw Error('The confirmations service is not connected')
+    const { name, status, ownerAddress } = filters
+
+    const results = await this.service.find({
+      query: {
+        placed: status === 'placed',
```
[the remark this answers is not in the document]
jurajpiar: We do. unfortunately, this is the result of having both "your domains" and "your offers" as filters. Had we the "your domains" show all the domains, including the placed ones, we could have the 'placed' as a filter toggle only.
This way, the status is defined as `export type DomainsSaleStatus = 'owned' | 'placed' | 'sold'`. This is partially why I would prefer moving to tabs, something like: "Buy domain, your domains, sold domains", where the placed would be a filter on "your domains" tab.

src/components/organisms/rns/sell/MyOffers.tsx
```diff
@@ -106,12 +70,10 @@ const MyOffers: FC<{}> = () => {
...
           handleSelect={() => {
             dispatch({
-              type: MARKET_ACTIONS.SELECT_ITEM,
+              type: 'SET_ORDER',
               payload: {
-                listingType: LISTING_TYPE,
                 item: domainItem,
-                txType: TxType.SELL,
-              },
+              } as any,
```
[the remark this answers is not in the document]
jurajpiar: Hmm... not sure, but I agree that we should handle these situations better, not just with casting to any.

src/components/organisms/rns/sell/SoldDomains.tsx
jurajpiar: Good point! This looks like an eye-patch on a split skull. It should definitely be removed and the itemCollection in the MarketPageTemplate properly typed. Right now, it is a collection of MarketItem which is clearly bollocks. We should extract interfaces from the headers (display item ones) and add them to a collective type somewhere that would be then used as type for the itemCollection.
That said, I don't think that should be in the scope of the store refactor. Instead, we should make an issue for that, probably.

src/components/pages/rns/buy/DomainOffersCheckoutPage.tsx
jurajpiar: Don't think that would work, as the null is returned so that the page renders that whilst redirecting. And component cannot return void. Only ReactElement | null.

src/components/pages/rns/buy/DomainOffersPage.tsx
jurajpiar: Same as in [SoldDomains](https://github.com/rsksmart/rif-marketplace-ui/pull/210#discussion_r451849250).

src/store/Market/rns/OffersStore.tsx
jurajpiar: Sure I can. I could also try naming it better, since their function is clearly not clear .]
As to extracting, yes, I pondered upon that (also domains store and sold store are pretty much identical), but didn't want to waste more time with that. But if you or anyone else insist I will dry it.

src/utils/typeUtils.ts
jurajpiar: found it online .)

src/store/storeUtils/reducer.ts
jurajpiar: Thanks man, I'm pretty happy with that myself.)

src/components/pages/rns/buy/DomainOffersCheckoutPage.tsx
farinalvaro: Nice, thanks for the explanation 🙂

julianmrodri:
Issue with Canceling Listed Domain: After you receive the final transactions it moves to an empty screen instead of getting the Confirmation (thumbs up screen)
image

julianmrodri:
Issue 2: There seems to be a different behavior on how the Domain listing is refreshed. In previous version if you were in the "Confirmation (thumbs up") page after Listing a domain and then you click "View Domain Listings" you would see the new domain populated (CORRECT BEHAVIOR).

In the current new version even if you do that and click on Domains you see the empty screen and the Refresh message which forces you ro refresh). I also noticed this with any change, looks like the Refresh is now enforced even when its not necessary.

Step to reproduce:
1) List a domain
2) IN confirmation scree click "View Domain listings"
3) it fill force you to refresh in the Domains list instead of just diplaying it


image

jurajpiar:
> Issue 2: There seems to be a different behavior on how the Domain listing is refreshed. In previous version if you were in the "Confirmation (thumbs up") page after Listing a domain and then you click "View Domain Listings" you would see the new domain populated (CORRECT BEHAVIOR).
> 
> In the current new version even if you do that and click on Domains you see the empty screen and the Refresh message which forces you ro refresh). I also noticed this with any change, looks like the Refresh is now enforced even when its not necessary.
> 
> Step to reproduce:
> 
> 1. List a domain
> 2. IN confirmation scree click "View Domain listings"
> 3. it fill force you to refresh in the Domains list instead of just diplaying it
>    Note: The same happend me right after I BUY a domain, and then click "Buy Another Domain" I still see the domain I just bought and the Refresh message to remove it.
> 
> image

yeap, that does seem to happen. not sure how I missed it. thanks Juli.

jurajpiar:
> Issue 1: with Canceling Listed Domain: After you receive the final transactions it moves to an empty screen instead of getting the Confirmation (thumbs up screen)
> image

Will fix

jurajpiar:
> Issue 1: with Canceling Listed Domain: After you receive the final transactions it moves to an empty screen instead of getting the Confirmation (thumbs up screen)
> image

Confirmed, will fix.

jurajpiar:
> Functionality issues detected will be reported in the comments.

all done now

src/components/organisms/rns/sell/MyOffers.tsx
```diff
@@ -1,82 +1,46 @@
-import IconButton from '@material-ui/core/IconButton'
+import { IconButton } from '@material-ui/core'
```
vojtechsimetka, line 1: Generally it's better to just import the IconButton itself as it will likely result in smaller bundle. That would however have to be the case everywhere in our code so thats a lost cause :)

src/components/organisms/rns/sell/SoldDomains.tsx
```diff
@@ -92,7 +47,7 @@ const SoldDomains: FC<{}> = () => {
       } = domainItem
       const currency = crypto[paymentToken]
 
-      const pseudoResolvedName = domainFilters?.name?.$like && `${domainFilters?.name?.$like}.rsk`
+      const pseudoResolvedName = !!name && `${name}.rsk`
```
vojtechsimetka, line 50:
```suggestion
      const pseudoResolvedName = name && `${name}.rsk`
```
The `!!` is unnecessary, the end result is the same

src/store/App/AppStore.tsx
```diff
@@ -33,10 +25,17 @@ export interface AppStoreProps {
 }
 
 export const initialState: AppState = {
-  apis: new Map([['confirmations', new ConfirmationsController()]]),
+  storeID: 'app',
+  apis: {
+    confirmations: new ConfirmationsController(),
+    offers: new OffersController() as APIController,
```
vojtechsimetka, line 31: Do you need to explicitly typecast it here? Can't you inherit or instantiate it as APIController?

julianmrodri:
@jurajpiar The fix to the refresh issue improved  but I don't think is the solution we want. It  still displays the Refresh message for a second and the change in the list is noticed by the user after clicking the page.   I think it should not be perceived at all the change after opening the page, it should be resolved before. Is it difficult or complex to achieve this behavior ?

src/api/models/RnsFilter.ts
```diff
@@ -1,28 +1,15 @@
...
-export interface DomainFilter extends MarketFilter {
-    ownerAddress: string
-    name?: {
-        $like: string
-    }
-    price?: {
-        $lte: number
-        $gte: number
-    }
+export type DomainsSaleStatus = 'owned' | 'placed' | 'sold'
```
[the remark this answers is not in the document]
jurajpiar: I have, in fact, changed all string enums to string literal types on purpose. The decision based on a number of articles about TS including the TS docs. String enums have bigger footprint and require more processing as they need to be transpiled into a lookup table. On the other hand string literal types are not represented in the transpiled code ([TS playground](https://www.typescriptlang.org/play/index.html#code/FAFwngDgpgBAygFQEoEkByBxA+gGRQgUSQEEc4YBeGAcjQHk6AFamAHxoDEUdCkX3qcAglwpE6DPxp0AqggAixQlMHCsdJPKIqkBDrrgAJFUJGMkdDAbgqAwjgLEk6zdrY0ZjRYVEBZfDbAwADGAPYAdgDOIDDBADZQAIYATnTJACZQyQBc8MgSoryk5FQARPaOzhpaSKVBwFDhAK4AtjAEaDK+5ADewDAw9EyUNEPMADT9MFw8RCPUM7zUkwOmouKY82t4G5IrMLIKSgTzh94Ey1Nr1XNUqiI3fPu6+gRG8y8Gxvtr5pbWWzUfysbxs+wqThcNXmEKqrieU085z8AXmSOOKIQgQAvkEwlEYgAzACWcRAWVyHS6JXanW6ADpFkQgA)), while having the enum-like property of being able to select it and not being able to pass anything that isn't an option in the literal type. Have a look at [string-enums](https://www.typescriptlang.org/docs/handbook/enums.html#string-enums) vs [string-literal-types](https://www.staging-typescript.org/docs/handbook/literal-types.html#string-literal-types)

If the answer does not satisfy you I'm happy to discuss it further :)

src/api/rif-marketplace-cache/rns/offers.ts
```diff
@@ -0,0 +1,65 @@
...
+    const results = await this.service.find({
+      query: {
+        domain: name ? {
+          name: {
+            $like: name,
+          },
+        } : undefined,
+        price: price ? {
+          $gte: price.min * 10 ** 18,
+          $lte: price.max * 10 ** 18,
```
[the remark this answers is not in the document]
jurajpiar: We have not defined the mechanism for different contracts yet. Like, when to use what multipliers, where should these be stored etc. Having said that I'm happy to improve it, just not sure this is in the scope of the PR. Perhaps create an issue?

src/components/Routes.tsx
```diff
@@ -42,8 +35,8 @@ const Routes = () => {
       <Route exact path={ROUTES.DOMAINS.CHECKOUT.BUY} component={DomainOffersCheckoutPage} />
       <Route exact path={ROUTES.DOMAINS.SELL} component={SellDomainsListPage} />
       <Route exact path={ROUTES.DOMAINS.CHECKOUT.SELL} component={DomainsCheckoutPage} />
-      <Route exact path={ROUTES.DOMAINS.CHECKOUT.CANCEL} component={CancelDomainCheckoutPage} />
       <Route exact path={ROUTES.DOMAINS.DONE.SELL} component={DomainListed} />
+      <Route exact path={ROUTES.DOMAINS.CHECKOUT.CANCEL} component={CancelDomainCheckoutPage} />
```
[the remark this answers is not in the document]
jurajpiar: Yes, but looks like I didn't finish it. Just wanted to make it a bit more clearer, ie group related ones. Perhaps grouping by tx type would be good:
```
      <Route exact path={ROUTES.DOMAINS.BUY} component={DomainOffersPage} />
      <Route exact path={ROUTES.DOMAINS.CHECKOUT.BUY} component={DomainOffersCheckoutPage} />
      <Route exact path={ROUTES.DOMAINS.DONE.BUY} component={DomainPurchased} />
      <Route exact path={ROUTES.DOMAINS.SELL} component={SellDomainsListPage} />
      <Route exact path={ROUTES.DOMAINS.CHECKOUT.SELL} component={DomainsCheckoutPage} />
      <Route exact path={ROUTES.DOMAINS.DONE.SELL} component={DomainListed} />
      <Route exact path={ROUTES.DOMAINS.CHECKOUT.CANCEL} component={CancelDomainCheckoutPage} />
      <Route exact path={ROUTES.DOMAINS.DONE.CANCEL} component={DomainCanceled} />
```
?

src/components/organisms/TransactionInProgressPanel.tsx
```diff
@@ -26,31 +27,32 @@ const useStyles = makeStyles((theme: Theme) => createStyles({
...
   const classes = useStyles()
 
   const { state: { confirmations: { txHash, currentCount, targetCount } }, dispatch: bcDispatch }: BlockchainStoreProps = useContext(BlockchainStore)
-  const { dispatch: mDispatch } = useContext(MarketStore)
 
   useEffect(() => {
     if (currentCount && targetCount && currentCount >= targetCount) {
       bcDispatch({
-        type: BLOCKCHAIN_ACTIONS.CLEAR_CONFIRMATIONS,
+        type: 'CLEAR_CONFIRMATIONS',
```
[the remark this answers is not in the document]
jurajpiar: https://github.com/rsksmart/rif-marketplace-ui/pull/210#discussion_r452694995

src/App.tsx
```diff
@@ -51,20 +54,26 @@ const App = () => {
...
-                      </Alert>
-                    </Collapse>
-                    <Routes />
-                  </PageTemplate>
-                  <Footer />
-                </div>
-              </Router>
+              <RnsDomainsStoreProvider>
+                <RnsOffersStoreProvider>
+                  <RnsSoldStoreProvider>
```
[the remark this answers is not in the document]
jurajpiar: Yes, this is something we need to improve. As mentioned in the description:
> all store providers are mounted at the start of the application. this should change to have been only mounted when needed
We need to figure out a mechanism for lazy provider mounting (that sounds dirty.]).

On the other hand, yes, until we do I could wrap them in `Array.reduce` that would return nested object with the app body being the leaf

src/api/rif-marketplace-cache/rns/offers.ts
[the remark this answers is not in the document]
vojtechsimetka: I agree we can do that in the next PR. I am just suggesting to do `toBaseValue(price.min, 18)` instead of `price.min * 10 ** 18`

src/components/Routes.tsx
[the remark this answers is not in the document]
vojtechsimetka: Sure, it just seemed very random

src/components/organisms/rns/sell/MyOffers.tsx
jurajpiar: Makes sense. And perhaps we should change it everywhere, but not in this this PR.

src/components/organisms/rns/sell/SoldDomains.tsx
jurajpiar: Well, while you are right about that the end result is the same, I thought it would be clearer if the constant held `false` instead of `''` if the string is empty.
It's not a very strong argument so, but by all means, I can change that if that didn't convince you.

src/api/rif-marketplace-cache/rns/offers.ts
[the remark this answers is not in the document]
jurajpiar: I agree

src/App.tsx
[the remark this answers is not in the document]
vojtechsimetka: I think that makes more sense

src/components/organisms/filters/DomainFilters.tsx
```diff
@@ -1,72 +1,59 @@
...
+        status: value,
       },
     })
   }
 
   return (
     <>
       <SearchFilter
-        value={searchValue}
+        value={name || ''}
```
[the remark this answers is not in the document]
jurajpiar: Well, this goes straight to the material ui `TextField` component, that unfortunately does not have `defaultValue` field.
I'd leave it here, as it prevents the `TextField` from crashing on `undefined`.
As an improvement of kind I could do `const searchValue = name || defaultValue` where `const defaultValue = ''`. Seems bit silly tho.
Another alternative is to pick that prop in the `SearchFilter` component and do the same there. Don't like either of the options very much tho.

src/store/App/AppStore.tsx
jurajpiar: Yes, it is quite ugly, but couldn't figure out the right way to do it. It already inherits from `APIController` but that defines fetch as 
```
fetch: (filters?: MarketFilter) => Promise<any>
```
whereas `OffersController` for example defines it as
```
fetch: (filters: Partial<RnsFilter>) => Promise<RnsDomainOffer[]>
```
where RnsFilter inherits from MarketFilter, but
> Type 'undefined' is not assignable to type 'Partial\<RnsFilter\>'.
as filters in general do not have to be passed to fetch (e.g. `ConfirmationsController`)

src/api/rif-marketplace-cache/rns/domains.ts
```diff
@@ -0,0 +1,51 @@
...
+    name,
+    tokenId,
+  }
+
+  if (offers?.length) {
+    const offer = offers[0]
+    domain.offer = {
+      ...offer,
+      paymentToken: getAvailableTokens[offer.paymentToken.toLowerCase()],
+      price: parseInt(offer.price, 10) / 10 ** 18,
```
nduchak, line 26: D we have some `amount-formater` util?

jurajpiar: not yet, but we should. But not in the scope of this PR. See https://github.com/rsksmart/rif-marketplace-ui/pull/210#discussion_r452697751

PR #219, opened 2020-07-01
Handle price and updates rif-ui dependency

### LINKED ISSUE #216: Prevent user from connecting with Multiple Wallets
We should prevent users from connecting their Wallets if we detect multiple `web3` wallets such as Nifty and Metamask, both enabled. If we cannot guarantee which wallet is open this could be dangerous for users and provide a bad UX.

We should notify the user to have only **one** web3 browser wallet enabled.

### LINKED ISSUE #217: Validations on Listing Price
We are currerntly not doing validations when entering the Price for a Domain in RIF. Then the transaction FAILS.

We should validate:
 * It is not empty and greater than zero
* No strange characters allowed, only numbers and decimal point.

### LINKED ISSUE #218: Price Filter errors
Reproduction steps:

Navigate to buy screen
Increase min or max amounts repeatedly by adding 9's or any number.	
Expected result:Number is increased to a defined maximum allowed.	
Actual result:Exponential numbers are shown, then Infinity and finally, NaN is shown and the filters get broken

farinalvaro:
This PR closes #217, closes #216 and closes #218  

Handles the price on before submitting a Domain for sell (#218).

Points to a [new version of `rif-ui`](https://github.com/rsksmart/rif-ui/pull/98) that fixes #216 and #217

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2989,8 +2989,9 @@
       }
     },
     "@rsksmart/rif-ui": {
-      "version": "github:rsksmart/rif-ui#b912e02f8bb2418a08fee88512be2956a662c565",
-      "from": "github:rsksmart/rif-ui#hotfix/sliderEdgeValues"
+      "version": "0.1.3",
+      "resolved": "https://registry.npmjs.org/@rsksmart/rif-ui/-/rif-ui-0.1.3.tgz",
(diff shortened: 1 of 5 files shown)


julianmrodri:
Approved. Dont forget to change the reference to RIF-UI once that is merged and released! Good Job!!

PR #250, opened 2020-07-14
Feature/error handling

jurajpiar:
Resolves [#RMKT191](https://rsklabs.atlassian.net/browse/RMKT-191)


src/store/App/appActions.ts
```diff
@@ -1,18 +1,25 @@
...
+}>
+
+export type ErrorMessagePayload = Modify<MessagePayload, ErrorMessage>
+
+export type RemoveMessagePayload = {
+  id: MessageId
 }
 
-export type AppPayload = LoadingPayloadType & MessagePayloadType
+export type AppPayload = LoadingPayload & (MessagePayload | ErrorMessagePayload) & RemoveMessagePayload
```
farinalvaro, line 22: I like this renaming 👌

src/store/Market/rns/DomainsStore.tsx
```diff
@@ -113,10 +116,18 @@ export const RnsDomainsStoreProvider = ({ children }) => {
...
           payload: { refresh: false },
-        } as any)
+        })
+      }).catch((e: Error) => {
+        appDispatch({
+          type: 'SET_MESSAGE',
+          payload: {
+            text: `Couldn't fetch RNS domains. Error: ${e.message}`,
+            type: 'error',
+          } as any, // TODO: any
```
farinalvaro, line 126: I think we are missing an id prop on this payload. With that, could we properly type it?

src/store/App/AppStore.tsx
```diff
@@ -1,22 +1,42 @@
-import { ServiceMap } from 'api/models/apiService'
+import { APIErrorId, ServiceMap } from 'api/models/apiService'
 import { ConfirmationsService } from 'api/rif-marketplace-cache/blockchain/confirmations'
+import { XRService } from 'api/rif-marketplace-cache/rates/xr'
 import { DomainsService } from 'api/rif-marketplace-cache/rns/domains'
 import { OffersService } from 'api/rif-marketplace-cache/rns/offers'
 import { SoldDomainsService } from 'api/rif-marketplace-cache/rns/sold'
+import { Severity } from 'components/molecules/InfoBar'
```
farinalvaro, line 7: If we are going to import and export this Material-UI prop, could we export it here and then reuse it in other components instead of exporting in the component? I just think it would be easier to remember where it lives :)

src/components/organisms/ErrorPanel.tsx
```diff
@@ -0,0 +1,49 @@
...
+            isVisible
+            buttonText={customAction ? customAction.name : 'dismiss'}
+            button={{
+              onClick: (event: React.MouseEvent<HTMLButtonElement, MouseEvent>) => {
+                if (customAction) {
+                  customAction.action(event)
+                }
+                dispatch({
+                  type: 'REMOVE_MESSAGE',
+                  payload: { id } as any, // TODO: any
```
farinalvaro, line 38: Could we use `RemoveMessagePayload` instead of `any`?

src/components/pages/rns/buy/DomainOffersCheckoutPage.tsx
```diff
@@ -108,20 +111,33 @@ const DomainOffersCheckoutPage: FC<{}> = () => {
...
-          const tokenPlacement = await marketPlaceContract.getPlacement(tokenId, { from: account })
-          const price = tokenPlacement[1]
+        const myBalance = await rifContract.getBalanceOf(account, { from: account })
+          .catch((error) => reportError({
+            error,
+            id: 'contract-rif-getBalanceOf',
+            text: 'Could not get balance from the crypto contract.',
+          }))
+
+        if (!myBalance) return
```
farinalvaro, line 121: Could we `return` after reporting the error and then report another error if there is no balance and return?
I'm just thinking if there could be a situation where it doesn't throw but there is no balance and we just `return undefined`

src/components/pages/rns/buy/DomainOffersCheckoutPage.tsx
```diff
@@ -180,39 +196,40 @@ const DomainOffersCheckoutPage: FC<{}> = () => {
...
-        bcDispatch({
-          type: 'SET_TX_HASH',
-          payload: {
-            txHash: transferReceipt.transactionHash,
-          } as AddTxPayload,
-        })
-        setIsPendingConfirm(true)
-      } catch (e) {
-        logger.error('Could not complete transaction:', e)
+      if (!transferReceipt) return
```
farinalvaro, line 224: Same situation as with `myBalance` above

src/components/pages/rns/cancel/CancelDomainCheckoutPage.tsx
```diff
@@ -145,39 +149,46 @@ const CancelDomainCheckoutPage = () => {
...
+      const rnsContract = RNSContract.getInstance(web3)
+      const marketPlaceContract = MarketplaceContract.getInstance(web3)
+
+      // Get gas price
+      const gasPrice = await web3.eth.getGasPrice()
+        .catch((error) => reportError({
+          error,
+          id: 'web3-getGasPrice',
+          text: 'Could not retreive gas price from the blockchain.',
+        }))
```
farinalvaro, line 161: Should we return if there was an error?

src/components/pages/rns/cancel/CancelDomainCheckoutPage.tsx
```diff
@@ -145,39 +149,46 @@ const CancelDomainCheckoutPage = () => {
...
+      // Send Unplacement transaction
+      const unplaceReceipt = await marketPlaceContract.unplace(tokenId, { from: account, gasPrice })
+        .catch((error) => reportError({
+          error,
+          id: 'contract-marketplace-unplace',
+          text: `Could not unplace domain ${name}.`,
+        }))
+      logger.info('unplaceReceipt:', unplaceReceipt)
+
+      if (!unplaceReceipt) return
```
farinalvaro, line 183: Same situation as with `DomainOffersCXheckoutPage`

src/components/pages/rns/sell/DomainCheckoutPage.tsx
```diff
@@ -147,39 +150,46 @@ const DomainsCheckoutPage: FC<{}> = () => {
...
+        .catch((error) => reportError({
+          error,
+          id: 'contract-rns-approve',
+          text: `Could not approve domain ${name}.`,
+        }))
+      logger.info('approveReceipt:', approveReceipt)
 
-        // Send Placement transaction
-        const placeReceipt = await marketPlaceContract.place(tokenId, rifTokenAddress, price, { from: account, gasPrice })
+      if (!approveReceipt) return
```
farinalvaro, line 174: Same situation about returning when we report an error that doesn't allow to continue with the process

src/components/templates/MarketPageTemplate.tsx
```diff
@@ -40,43 +44,57 @@ const MarketPageTemplate: FC<MarketPageTemplateProps> = ({
...
+        payload: {
+          type: 'warning',
+          text: 'Please, connect your wallet.',
+          id: 'wallet',
+          // customAction: {
+          //   name: 'Connect',
+          //   action: () => {
+          //     return <Login /> //TODO: dispatch web3dispatch action 'CONNECT' that would trigger the wallet connection modal to open
+          //   }
+          // }
```
farinalvaro, line 70: May we remove this comment?

src/contracts/interfaces.ts
```diff
@@ -0,0 +1,36 @@
...
+//     errorReporter!: ErrorReporter
+//     web3!: Web3
+
+//     public static getInstance(web3: Web3, errorReporter: ErrorReporter): Contract {
+//         if (!AbstractContract.instance) {
+//             // AbstractContract.instance = new AbstractContract(web3)
+//         }
+//         return AbstractContract.instance
+//     }
+
```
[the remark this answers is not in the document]
jurajpiar: Ah I thought I did remove that. Well, will do.

src/store/App/AppStore.tsx
jurajpiar: Not sure I like the idea having material-ui imports in the store. 
On the other hand, you are right, it's not obvious why it lives in the InfoBar. Perhaps I can put it in the src/models?

src/components/pages/rns/buy/DomainOffersCheckoutPage.tsx
jurajpiar: observe that it actually returns the reportError. This is because that in turn returns void, thus rendering myBalance falsy.

src/components/pages/rns/cancel/CancelDomainCheckoutPage.tsx
jurajpiar: observe that it actually returns the reportError. This is because that in turn returns void, thus rendering myBalance falsy.

src/components/pages/rns/cancel/CancelDomainCheckoutPage.tsx
jurajpiar: observe that it actually returns the reportError. This is because that in turn returns void, thus rendering myBalance falsy.

src/components/templates/MarketPageTemplate.tsx
jurajpiar: I left it there as a reminder for us that we need to do that.

jurajpiar: but if you insist I can remove it :)

src/components/pages/rns/buy/DomainOffersCheckoutPage.tsx
jurajpiar: observe that it actually returns the reportError. This is because that in turn returns void, thus rendering myBalance falsy.

src/components/pages/rns/sell/DomainCheckoutPage.tsx
jurajpiar: observe that it actually returns the reportError. This is because that in turn returns void, thus rendering myBalance falsy.

src/store/App/AppStore.tsx
farinalvaro: You're right about it's weird to import Material-UI on the store.
But it's what we are doing indirectly. I like more your proposal of putting it in `src/models` 🙂

src/api/models/apiService.ts
```diff
@@ -1,34 +1,87 @@
...
+  fetch = async (filters?: MarketFilterType | any): Promise<any> => {
+    if (!this.service) {
+      this.errorReporter({
+        id: 'service-connection',
+        text: 'Error while fetching data.',
+        error: new Error('The confirmations service is not connected'),
+      })
+    }
+    let data = []
+    try {
```
farinalvaro, line 46: May we move this class to a new folder like `src/services`?

jurajpiar: seems unnecessary to add a folder just for one class and an interface, no?

farinalvaro: Fair enough 🙂 we could do it if it grows

julianmrodri:
When testing this seems the UI is not waiting for the Required confirmations. When placing or canceling I get the message of "Waiting for first confirmation" but right after that I ge thr confirmation page even though I have Confirmations = 2 in the Cache and I havent forwarded  the blockchain.

julianmrodri:
I would fix the Warning and its OK!! Good job

PR #279, opened 2020-07-29
Storage listing

farinalvaro:
This PR solves [RMKT-132](https://rsklabs.atlassian.net/browse/RMKT-132) and [RMKT-135](https://rsklabs.atlassian.net/browse/RMKT-135)


src/components/molecules/CombinedPriceCell.tsx
```diff
@@ -13,11 +14,11 @@ export interface CombinedPriceCellProps {
 const CombinedPriceCell: FC<CombinedPriceCellProps> = ({
   className = '', divider, price, currency, priceFiat, currencyFiat,
 }) => {
-  const cappedDecimalsFiat = parseFloat(priceFiat).toFixed(4).toString()
+  const cappedDecimalsFiat = priceDisplay(parseFloat(priceFiat), 4)
 
   return (
     <div className={(`priceCell ${className}`).trim()}>
-      <PriceItem key="hola" type="crypto" price={price} currency={currency} />
```
jurajpiar, line 20: lol well spotted

src/utils/parsers.ts
```diff
@@ -1,14 +1,26 @@
...
-  return big.times(10 ** decimalPlaces)
-    .toString()
+  return big.times(10 ** decimalPlaces).toString()
+}
+
+// parses to wei ignoring the decimals
+export const parseToWei = (src: BigSource): string => {
+  const weiString = parseToBigDecimal(src, -18).toString()
+
+  return weiString.split('.')[0]
```
jurajpiar, line 25:
```suggestion
  return parseInt(parseToBigDecimal(src, -18)).toString()
```
or
```suggestion
  return Math.floor(parseToBigDecimal(src, -18)).toString()
```

src/utils/utils.ts
```diff
@@ -1,5 +1,16 @@
 export const mayBePluralize = (count, noun, suffix = 's'): string => `${count} ${noun}${count !== 1 ? suffix : ''}`
 
+export const priceDisplay = (value: number, maxDecimals = 8): string => value
+  .toFixed(maxDecimals)
+  .toString()
+  .replace(/[.,]00000000$/, '')
```
jurajpiar, line 6: what is this suppose to do? Why are the zeros hardcoded? Can this be done better?

src/utils/parsers.ts
farinalvaro: Nice suggestion, thanks 🙂

src/utils/utils.ts
```diff
@@ -1,5 +1,16 @@
...
+
+export const convertGbsToBytes = (gbs: number): number => gbs * 1024 * 1024 * 1024
```
jurajpiar, line 8: 1024 * 1024 * 1024 === 1024 ** 3

src/utils/utils.ts
farinalvaro: Good point, this aims to remove the unnecessary zeroes after the `,|.` that the `toFixed` adds. Will improve it 🙂

src/utils/utils.ts
farinalvaro: You are right about the `**`. About making it more general... meh 😅

PR #306, opened 2020-08-14
feat: adds storage offers listing

jurajpiar:
Adds Storage offers browsing

Resolves [RMKT-13](https://rsklabs.atlassian.net/browse/RMKT-13)

As a result of merging master in, this PR has to be squashed into master.


src/components/atoms/ItemWUnit.tsx
```diff
@@ -0,0 +1,30 @@
...
+    color: colors.gray3,
+    fontSize: fonts.size.normal,
+  },
+}
+
+interface StylesProps {
+  classes: Record<keyof typeof styles, string>
+}
+
+export interface ItemWUnitProps extends StylesProps {
```
farinalvaro, line 20: Can we call it `ItemWithUnit` 😄?

src/api/rif-marketplace-cache/storage/offers.ts
```diff
@@ -0,0 +1,55 @@
...
+      })),
+    pricePGBPDay: plans
+      .reduce<Big>((acc, plan) => {
+        const period = new Big(plan.period)
+        const price = parseToBigDecimal(plan.price, 18)
+        const combinedPrice = acc.add(price.div(period))
+        return combinedPrice
+      }, new Big(0))
+      .div(plans.length)
+      .mul(new Big(PeriodInSeconds.Daily)),
```
farinalvaro, line 40: perfect 👌

src/components/templates/MarketPageTemplate.tsx
```diff
@@ -21,6 +21,7 @@ export interface MarketPageTemplateProps {
   requiresAccount?: boolean
   dispatch: Dispatch<RnsAction>
   outdatedCt: number
+  heading?: React.ElementType
```
farinalvaro, line 24: Just for consistency, should we keep the `Heading` prop starting with uppercase? here too

src/components/pages/storage/StorageRoutes.tsx
```diff
@@ -1,14 +1,16 @@
 /* eslint-disable-next-line import/no-unresolved */
 import { StyledNavTabProps } from '@rsksmart/rif-ui/dist/components/atoms/StyledNavTab'
-import React, { FC } from 'react'
-import ROUTES from 'routes'
+import React, { useEffect } from 'react'
 import {
-  Redirect, Route, Switch, useLocation,
+  Redirect, Route, Switch, useHistory, useLocation,
```
farinalvaro, line 5: Do we need both `useLocation` and `useHistory`?

src/components/pages/storage/buy/__tests__/StorageOffersPage.spec.tsx
```diff
@@ -0,0 +1,146 @@
+// const FAKE_BILLING_PLAN: BillingPlan[] = [
```
farinalvaro, line 1: I know this is valuable, but I'm afraid this commented code could stay here forever 😄

src/components/pages/storage/buy/StorageOffersPage.tsx
```diff
@@ -1,72 +1,88 @@
...
-          <Marketplace items={itemCollection} headers={headers} isLoading={false} />
-        </Grid>
-      </>
-    </Grid>
+    <MarketPageTemplate
+      className="Storage Offers"
+      filterItems={<StorageFilters />}
+      itemCollection={itemCollection}
+      headers={TABLE_HEADERS}
+      dispatch={dispatch as any} // FIXME: Change the type in the MarketPageTemplate
```
farinalvaro, line 82: Out of the scope of this PR, but what do you think if instead of passing the dispatch we pass the action we want to perform? Something less generic. If it's ok for you I could open an issue with that

src/components/pages/storage/buy/StorageOffersPage.tsx
```diff
@@ -1,72 +1,88 @@
...
-    provider: 'Provider',
-    location: 'Location',
-    system: 'System',
-    availableSize: 'Available Size',
-    subscriptionOptions: 'Subscription Options',
-    pricePerGb: 'Price/GB',
-    action1: '',
-  }
-  const itemCollection = []
+  // const heading = () => <Typography>Result</Typography>
```
farinalvaro, line 74: Can we uncomment this?

src/routes.ts
```diff
@@ -1,55 +1,55 @@
...
-const DOMAIN_OFFERS_DONE = `${DOMAINS_BUY}/done`
-const DOMAINS_DONE = `${DOMAINS_SELL}/done`
-const DOMAIN_CANCEL_CHECKOUT = `${DOMAINS_SELL}/cancel/checkout`
-const DOMAIN_CANCEL_DONE = `${DOMAINS_SELL}/cancel/done`
+
+const RNS = '/domains'
+const RNS_BUY = `${RNS}/buy`
+const RNS_SELL = `${RNS}/sell`
+const RNS_CANCEL = `${RNS_SELL}/cancel`
+
```
farinalvaro, line 11: Much cleaner, nice!

src/context/Services/storage/storageActions.ts
```diff
@@ -0,0 +1,37 @@
...
+  NOOP: (state, payload) => ({ ...state, ...payload }),
+  SET_LISTING: (state, payload) => ({
+    ...state,
+    listing: {
+      items: payload.items,
+      outdatedTokens: [],
+    },
+  }),
+  SET_ORDER: (state, payload) => ({ ...state, ...payload }),
+}
```
farinalvaro, line 37: Is this reducer used for the current order or more generic one?
If it's only for the order, why do we reference the listing here?

src/api/models/storage/transports.ts
```diff
@@ -0,0 +1,19 @@
...
+    id: number
+    period: string
+    price: string
+    offerId: string
+    createdAt: string
+    updatedAt: string
+}
+
+export interface OfferTransport {
+    peerId: string | null
```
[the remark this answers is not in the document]
jurajpiar: no because that means `undefined` which isn't the same as `null`

src/components/pages/storage/buy/StorageOffersPage.tsx
jurajpiar: Just so that you know I'm not making this up:
> useReducer also lets you optimize performance for components that trigger deep updates because you can pass dispatch down instead of callbacks. https://reactjs.org/docs/hooks-reference.html#usereducer

src/context/Services/storage/storageActions.ts
jurajpiar: you lost me with this. the order has nothing to do with the listing. technically, the SET_ORDER action shouldn't even be in this PR.

src/components/atoms/ItemWUnit.tsx
jurajpiar: Could you explain why?
I think it is obvious for anyone who speaks english what that W stands for no?
Like you don't question what say, "steak w/ salad" means, right? Equally "item w/ unit" seems clear to me.
But I have no real quarrel with changing it if you would prefer it that way.

src/context/Services/storage/storageActions.ts
farinalvaro: You're right, my fault, I got confused with the `listing` word... again 😅

src/components/atoms/ItemWUnit.tsx
farinalvaro: Fair enough 🙂

src/components/pages/storage/buy/__tests__/StorageOffersPage.spec.tsx
jurajpiar: Probably wouldn't since it clearly needs to be done, but I will remove it

src/components/pages/storage/buy/StorageOffersPage.tsx
jurajpiar: No, actually I'm going to remove it altogether.

src/components/templates/MarketPageTemplate.tsx
jurajpiar: Yea probably.

PR #319, opened 2020-08-21
feat: rns enhancements

julianmrodri:
* Restore support for new column `id` for supporting multiple actions in one transaction.
* Update new RNS contract.
* Format messaging and handling additional validation if Token approved/not approved.


src/components/pages/rns/buy/DomainOffersCheckoutPage.tsx
```diff
@@ -246,6 +248,27 @@ const DomainOffersCheckoutPage: FC<{}> = () => {
...
+              text: `Could not retrieve approval for ${domainName} from contract.`,
+            })
+          })
+
+        // Check if domain is approved
+        const isApproved = tokenApproval[1]?.toLowerCase() === marketPlaceAddress
+
+        if (!isApproved) {
+          throw new UIError({
+            error: new Error(),
```
farinalvaro, line 266: I would suggest to add a description to the error so we can easily find it by reading the logs

PR #328, opened 2020-08-25
hotfix: fixes routing issue

jurajpiar:
- removes withTabs as it was instantiating new context on history change


julianmrodri (dismissed):
LGTM

PR #329, opened 2020-08-26
fix: better types and collects marketplace props

jurajpiar:
- removes casting to any from item prop assignment


farinalvaro (dismissed):
LGTM :)

julianmrodri (approved):
LGTM

farinalvaro (approved):

PR #353, opened 2020-09-08
New storage sell page

farinalvaro:
This PR adapts the storage sell section to the new format.

### Features
- Adds new styles for the page, organizes the steps with a vertical stepper
- Adds `PeerId` field on the form ~and storage contract wrapper~
- Removes `currency` field from the offer
- Adds currency per plan. Now we allow plans with the same `timePeriod` but different `currency`

### Areas for improvement
 - PeerId - as pointed in the comments, we need to create a proper mechanism to send the PeerId field to the contract
 - Tests - would be ideal to cover the validations of the offer and the plans, but as the validations are happening on the view, it's not possible to test directly on the reducer with unit tests.
 - The currency is being handled as a string everywhere. I think it would be best if we define a type for it so we stop propagating that. May be open an issue for that?
- Remove `TimePeriodEnum`, I think it should be removed on a different PR.


src/contracts/Storage.ts
```diff
@@ -47,16 +48,21 @@ class StorageContract {
...
-
-    return this.contract.methods
-      .setOffer(capacityMB, billingPeriods, billingRbtcWeiPrices, message)
-      .send({ from, gas, gasPrice }, (err, txHash) => {
-        if (err) return Promise.reject(err)
-        return waitForReceipt(txHash, this.web3)
-      })
+
+    const { asciiToHex, padRight } = this.web3.utils
+    const message = [padRight(asciiToHex(peerId), 64)]
```
AuHau, line 56: Unfortunately, this won't work.
"Messages" is a general concept in the Storage SC that allows passing information. As Rinke designed this, he defined that each message type is prefixed with one byte that defines what type of message it is. As you can see [here](https://github.com/rsksmart/rif-marketplace-storage#message) we currently have only one type and that is announcing PeerId type.
I originally was thinking about refactoring it into making the PeerId required for the Offer (see the [PR](https://github.com/rsksmart/rif-marketplace-storage/pull/105), but then we decided to not do it as it is not really mandatory from the SC point of view, but rather the system itself, so we gonna enforce that on Cache level...

So there are two problems here:
 - the message has to have `0x01` prefix
 - as the parameter is `byte32[]` and PeerId is usually 47 chars, you have to split the string into chunks. Btw. you should not really on the length 47 chars, make it generic solution.

For the splitting part you can see it working at [here](https://github.com/rsksmart/rif-storage-pinner/blob/63f2e8679a3e2900639f8fbbc9479a147ba35b80/test/utils.ts#L40), but... That is unfortunately not the full solution here (but can be used for encoding the IPFS hash, that is passed to the SC when you create new Agreement).

Problem is that when you call it as like this `const msg = encodeHash(peerId)` you get an array like `[first32bytes, second32bytes, ...]`, but then you should prefix the first element with the `0x01` which will make it overflow the `byte32` length. So there has to be "shifting" mechanism that moves the excess bytes into later elements, but I still have not got around to write this part, sooo maybe you will and I can steal it afterwards 😅 😇

farinalvaro: Hey Adam, thanks for this valuable comment 🙂
We were discussing with @jurajpiar and will be working on the PeerId on a separated PR, so I would better keep sending an empty message for now.
As for the shifting mechanism, let's talk and tackle it on that new PR

src/components/organisms/storage/sell/EditablePlanItem.tsx
```diff
@@ -67,65 +68,87 @@ const EditablePlanItem: FC<EditablePlanItemProps> = ({
...
-        price={pricePerGb}
-        currency={currency}
-        fiatPrice={fiatPrice}
-        fiatDisplayName={fiatDisplayName}
-        periodOptions={allPeriods}
-        selectedPeriod={timePeriod}
-        availablePeriods={availablePeriods}
+  const ActionButton = () => (editMode
+    ? (
+      <TooltipIconButton
```
jurajpiar, line 73: I don't think this needs to be a function here.

src/components/organisms/storage/sell/EditablePlanItem.tsx
```diff
@@ -67,65 +68,87 @@ const EditablePlanItem: FC<EditablePlanItemProps> = ({
...
+        tooltipTitle="Save plan"
+        icon={<SaveIcon />}
+        iconButtonProps={{
+          disabled: pricePerGb <= 0 || (
+            (
+              // the period or currency have changed and the selected option is in use
+              planItem?.timePeriod !== timePeriod || planItem?.currency !== currency
+            )
+            && usedPeriodsPerCurrency[currency].includes(timePeriod)
+          ),
```
jurajpiar, line 83: Nesting ternary operators is not encouraged. They are harder to read (as you have to keep the previous parts in mind), harder to debug and is prone to errors (similar to switch/case or even nested ifs). We seem to have that rule overwritten to just a warning but we should still avoid it where possible.

jurajpiar (changes requested):
Nice, just need to remove some smelly code .)

src/components/organisms/storage/sell/EditablePlanItem.tsx
farinalvaro: Thanks, good point

src/components/organisms/storage/sell/SellStepper.tsx
```diff
@@ -0,0 +1,86 @@
...
+import StepLabel from '@material-ui/core/StepLabel'
+import StepContent from '@material-ui/core/StepContent'
+import { Button } from '@rsksmart/rif-ui'
+import { StorageListingContextProps } from 'context/Services/storage/interfaces'
+import StorageListingContext from 'context/Services/storage/ListingContext'
+import GeneralFeatures from './GeneralFeatures'
+import PlanItemsList from './PlanItemsList'
+
+export interface SellStepperProps {
+  finalAction: ReactElement
```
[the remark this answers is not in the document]
farinalvaro: As it's a ReactElement, I followed your previous suggestion of renaming it to `endHandler`

src/context/Services/storage/listingReducer.ts
```diff
@@ -1,51 +1,64 @@
...
+): Record<string, []> => {
+  const usedPeriodsPerCurrency = {}
+  planItems.forEach((planItem) => {
+    usedPeriodsPerCurrency[planItem.currency] = [
+      ...(usedPeriodsPerCurrency[planItem.currency] || []),
+      planItem.timePeriod,
+    ]
+  })
+  return usedPeriodsPerCurrency
+}
```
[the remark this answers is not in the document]
farinalvaro: Done. Thanks

src/components/organisms/storage/sell/EditablePlanItem.tsx
farinalvaro: I think it just helps to avoid using the nested ternary operator as you suggested and also helps with the readability. Please, take a look at the new function without the ternary operator and let me know if you still think the same 🙂

jurajpiar (approved):
LGTM

Issue #384, opened 2020-10-02
Pagination, Sorting and New Branding RNS


julianmrodri:
We need to include in RNS (`release/mainnet` branch) Pagination, Sorting and the new branding.

julianmrodri:
Closing as most of these items were addressed, and will create separate issue for Sorting.

PR #389, opened 2020-10-05
feat(rns): pagination for mainnet release

jurajpiar:
Adds pagination

Contributes towards #384 
Works in tandem with https://github.com/rsksmart/rif-marketplace-cache/pull/320

### DIFF
--- a/src/api/rif-marketplace-cache/rates/xr.ts
+++ b/src/api/rif-marketplace-cache/rates/xr.ts
@@ -1,12 +1,15 @@
-import { AbstractAPIService, APIService } from 'api/models/apiService'
+import { Paginated } from '@feathersjs/feathers'
+import { AbstractAPIService, APIService, isResultPaginated } from 'api/models/apiService'
 import { Modify } from 'utils/typeUtils'
 
 export type XRServiceAddress = 'rates/v0'
 export const xeServiceAddress: XRServiceAddress = 'rates/v0'
 
 export type SupportedFiat = 'usd' | 'eur' | 'btc' | 'ars' | 'cny' | 'krw' | 'jpy';
-export type SupportedTokens = 'rif';
-export const tokenDisplayNames: Record<SupportedTokens, string> = {
+
+export const SUPPORTED_TOKENS = ['rif'] as const
+export type SupportedToken = typeof SUPPORTED_TOKENS[number];
+export const tokenDisplayNames: Record<SupportedToken, string> = {
   rif: 'RIF',
   // rbtc: 'RBTC',
 }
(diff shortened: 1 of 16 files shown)


src/api/rif-marketplace-cache/rates/xr.ts
```diff
@@ -17,21 +20,34 @@ export interface XRFilter {
...
         $select: ['token', fiatSymbol],
       },
-    }) as unknown as []
+    })
+    const { data, ...metadata } = isResultPaginated(results)
+      ? results : { data: results }
+    this.meta = metadata
 
-    return results
+    return data.filter((rate) => isSupportedToken(rate.token))
```
farinalvaro, line 51: `data.filter(({token}) => isSupportedToken(token))`
could be cleaner

farinalvaro (commented):
Looking good. Just left a minor comment

jurajpiar:
> Looking good. Just left a minor comment.
> Also, the commits are duplicated

This is going to get squashed.

farinalvaro (approved):
LGTM

julianmrodri:
Looks good overal!! a few issues:
1) when you are in page 3 for example and you reduce the filter, you end up with an empty page and need to navigate back to see the results. Would be nice to never end up in empty and in that case to get redirected to the first page.
2) When you change the filter we are not updating the index on the top right (like X / Y) but if you move to Page 2 for example it gets triggered the update. So Ideally we want this to be updated together with the new fetching of data based on the filter.

src/components/molecules/RifPaging.tsx
```diff
@@ -0,0 +1,59 @@
+import { Button, Grid, makeStyles } from '@material-ui/core'
+import { NavigateBefore, NavigateNext } from '@material-ui/icons'
+import React from 'react'
+
+const useButtonStyle = makeStyles(() => ({
+  root: {
+    minWidth: '1em',
+    // borderRadius: '50px',
```
farinalvaro, line 8: remove comment

src/components/molecules/RifPaging.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+const RifPaging = ({
+  from, to, total, onNext, onPrev,
+}) => {
+  const buttonStyleClasses = useButtonStyle()
+  return (
+    <Grid
+      container
+      direction="row"
+      spacing={0}
```
farinalvaro, line 24: I think spacing is 0 by default

src/components/molecules/RifPaging.tsx
```diff
@@ -0,0 +1,59 @@
...
+        </Button>
+      </Grid>
+      <Grid item>
+        { `${from}-${to}/${total}`}
+      </Grid>
+      <Grid item>
+        <Button
+          variant="text"
+          color="primary"
+        //   endIcon={<NavigateNext />}
```
farinalvaro, line 47: remove comment

src/components/molecules/RifPaging.tsx
```diff
@@ -0,0 +1,59 @@
...
+      container
+      direction="row"
+      spacing={0}
+      wrap="nowrap"
+    >
+      <Grid item>
+        <Button
+          variant="text"
+          color="primary"
+        //   startIcon={}
```
farinalvaro, line 31: remove comment

src/api/rif-marketplace-cache/rates/xr.ts
```diff
@@ -1,12 +1,15 @@
...
 
 export type XRServiceAddress = 'rates/v0'
 export const xeServiceAddress: XRServiceAddress = 'rates/v0'
 
 export type SupportedFiat = 'usd' | 'eur' | 'btc' | 'ars' | 'cny' | 'krw' | 'jpy';
-export type SupportedTokens = 'rif';
-export const tokenDisplayNames: Record<SupportedTokens, string> = {
+
+export const SUPPORTED_TOKENS = ['rif'] as const
+export type SupportedToken = typeof SUPPORTED_TOKENS[number];
```
farinalvaro, line 11: Why did you change this?

farinalvaro (commented):
Looking good 🙂 just left some comments on minor details.
I'm wondering if there's a way to update the pagination to a default state any time the items or filters change, may be adding an action to refresh the pagination and dispatch it on those cases?

src/api/rif-marketplace-cache/rates/xr.ts
jurajpiar: Need it as an array I can check agains in runtime in https://github.com/rsksmart/rif-marketplace-ui/pull/389/files#diff-2ce228329ef532bb41784984f09fff23R34.

farinalvaro (approved):

PR #405, opened 2020-10-13
Hotfix/sorting

jurajpiar:
Adds Sorting ability to rns offers


julianmrodri:
@jurajpiar we might probably want to close this PR and start a new one at some point (if we resume this feature) by branching directly from `master` right?

PR #420, opened 2020-10-22
fix: cleans up context on offers page and my offers will unmount

farinalvaro:
Cleans up the offers context on storage offers will unmount and fixes the missing dependencies array that I forgot to add on my previous PR (sorry 😅 )


julianmrodri (approved):
LGTM

PR #479, opened 2020-11-11
fix(storage): calculates withdrawable funds and disables the action when no funds

### LINKED ISSUE #470: Problems with Withdrawable and Available Funds
There is a problem with the values we are displaying for **Withdrawable** and **Available** funds for both Consumer and Provider. These are obtained from the Cache but they do not get updated unless there is transaction, so they become invalid and obsolete as time passes.

We need to get this values from the Contract, or ensure the Cache tracks offers/agreements and maintains an updated status in real time.

farinalvaro:
Closes #470 

Sometimes we are getting negative periods from the cache, https://github.com/rsksmart/rif-marketplace-cache/pull/384 would fix that

### DIFF
--- a/src/components/molecules/SelectRowButton.tsx
+++ b/src/components/molecules/SelectRowButton.tsx
@@ -1,7 +1,9 @@
 import React, { FC } from 'react'
 import { Button } from '@rsksmart/rif-ui'
+/* eslint-disable-next-line import/no-unresolved */
+import { ButtonProps } from '@rsksmart/rif-ui/dist/components/atoms/Button'
 
-export interface SelectRowButtonProps {
+export interface SelectRowButtonProps extends ButtonProps {
   className?: string
   id: string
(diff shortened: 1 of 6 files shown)


src/components/molecules/SelectRowButton.tsx
```diff
@@ -5,6 +5,7 @@ export interface SelectRowButtonProps {
   className?: string
   id: string
   handleSelect: (event: React.MouseEvent<HTMLButtonElement, MouseEvent>) => void
+  disabled?: boolean
 }
```
jurajpiar, line 9: on hindside, the `SelectRowButtonProps` should probably just expose MUI `ButtonProps`. or perhaps it shouldn't be there at all.

jurajpiar (approved):

julianmrodri (approved):
LGTM

PR #480, opened 2020-11-12
fix(notifications): filters by account

### LINKED ISSUE #478: Notifications being shown for incorrect user
I have noticed that if you are logged in with any user you can see the notifications in the DB. It seems it is not filtering properly by account. In this case im logged in with account 0xFF... but its displaying to me notifications which are only created for user 0x90...

image

Notifications Table:
image

jurajpiar:
Filters notifications by account.

Resolves #478 
Closes [RMKT-383](https://rsklabs.atlassian.net/jira/software/projects/RMKT/boards/60?assignee=5e84f3015b08780c1c4127b0&selectedIssue=RMKT-383)

### DIFF
--- a/src/api/rif-marketplace-cache/notifications/index.ts
+++ b/src/api/rif-marketplace-cache/notifications/index.ts
@@ -21,13 +21,26 @@ export const mapFromTransport: MapFromTransport<
   },
 })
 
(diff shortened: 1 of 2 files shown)


farinalvaro (approved):
LGTM

PR #481, opened 2020-11-12
fix: enables CopyTooltip in addresses and uses agreementReference

### LINKED ISSUE #466: Display DataReference instead of Agreement Reference
In the Agreement view detail for both Consumer and Provider we are showing in the HASH value the `agreementReference` instead of the `dataReference`. We should display the dataReference with is the real HASH of the pinned content.

image

### LINKED ISSUE #467: Allow Tooltip and copy for all addresses
Everytime we display and address we should show the Tooltip to display the COMPLETE address and the option to click and copy the value. This was already implemented in RNS pages we need to bring it to Storage.

farinalvaro:
- Puts all addresses with copy tooltip again - closes #467 
	- That copy tooltip component is looking awful in some sections, would be good to redesign it
- shows agreementReference instead of agreementsHash - closes #466


julianmrodri (approved):
LGTM

julianmrodri:
@jurajpiar  can you review so we merge it for staging version? Thanks!

PR #487, opened 2020-11-13
feat(storage): validates CID hash and prepends /ipfs/

### LINKED ISSUE #468: Validate IPFS Hash to Pin
Currently the UI allows any text to be provided as **Hash to pin**. The problem is that this text then gets converted to a hash format which may return an arbitrary values and can cause issues. 

One nice enhancement would be to validate that this is actually a hash in the expected format. Would be nice to device a mechanisms to validate the Hash provided.

jurajpiar:
What the tin says.

Resolves #468

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "rif-marketplace-ui",
-  "version": "1.0.0",
+  "version": "1.1.0-dev.1",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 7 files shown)


jurajpiar:
Ok, so... the CID lib makes our test unhappy
```
FAIL  src/App.test.tsx
  ● Test suite failed to run

    TypeError: TextDecoder is not a constructor

    > 1 | import CID from 'cids'
        | ^
      2 | 
      3 | export const isEmpty = (
      4 |   text: string | unknown,
...
```

There is an open issue at jest about this [here](https://github.com/facebook/jest/issues/9983). The multibase lib have informed react-scripts about it [(presumably)](https://github.com/multiformats/js-multibase/issues/66#issuecomment-677874507).
The "fix" as stated [there](https://github.com/Uniswap/uniswap-interface/pull/1056/files) is basically to upgrade react-script to v4.
Another [patch](https://stackoverflow.com/questions/57712235/referenceerror-textencoder-is-not-defined-when-running-react-scripts-test/57713960#57713960) I found is to create custom test environment.

julianmrodri:
We can comment the test and resume work later on this once we upgrade to `react 4` after release. But I dont think we should do that now. Also please consider that we also need to add the prefix `/ipfs/` to the hash using UPLOAD SERVICE. Currently it is not working as Pinner rejects anything without that `prefix`.

julianmrodri (changes requested):
* Comment failing test for now until we upgrade to React 4 (we know we have to do some analysis for this specially around `tasegir`).
* Add /ipfs/ also to hash created by Upload File Service

itofarina (commented):
LG! I agree with @julianmrodri about re-taking the failing test once we move to react-scripts v4

julianmrodri (dismissed):
LGTM

julianmrodri (approved):
LGTM 🚀

PR #490, opened 2020-11-16
chore: updates rif-ui dependency

### LINKED ISSUE #457: Slider
Anytime using the slider we should update the filters with `onChangeCommitted` instead of using `onChange`. That way the filters will not update until the `mouseup` event is triggered.

Reference: https://material-ui.com/api/slider/#props

farinalvaro:
Closes #457. The Slider of the new version of `rif-ui` will trigger changes only on committed.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "rif-marketplace-ui",
-  "version": "1.0.0",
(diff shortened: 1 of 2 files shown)


julianmrodri (approved):
LGTM

jurajpiar (approved):

PR #492, opened 2020-11-16
feat(storage): adds ability to check for whitelisted storage providers

### LINKED ISSUE #454: Validate Whitelisted Providers in Storage-Sell
In Storage sell we should notify the users if they cannot create an Offer. We should show the message

"This account is not registered as Storage Provider."  

And also disable the **"LIST STORAGE"** button.

Currently we are allowing the user to send the TX which fails:

image

farinalvaro:
## Features
- Adds ability to check for whitelisted storage providers and disables related actions - Closes #454 
- Adds storage global context to handle shared storage props (in a future refactor we could handle the contract instance here)`


src/components/pages/storage/StorageRoutes.tsx
```diff
@@ -83,88 +84,90 @@ const StorageRoutes: FC = () => {
...
-                  />
-                  <Route
-                    exact
-                    path={ROUTES.STORAGE.MYOFFERS.EDIT.BASE}
-                    component={StorageEditOfferPage}
-                  />
-                </Switch>
-              </OfferEditContextProvider>
+          <StorageContextProvider>
+            <Switch>
```
farinalvaro, line 88: Just to avoid confusion, the only change on this file was put the `StorageContextProvider` under `TabsTemplate`, all the other changes seem to be because of indentation

jurajpiar: thanks for that clarification.

src/components/pages/storage/StorageRoutes.tsx
```diff
@@ -83,88 +84,90 @@ const StorageRoutes: FC = () => {
...
-                    component={StorageMyOffersPage}
-                  />
-                  <Route
-                    exact
-                    path={ROUTES.STORAGE.MYOFFERS.EDIT.BASE}
-                    component={StorageEditOfferPage}
-                  />
-                </Switch>
-              </OfferEditContextProvider>
+          <StorageContextProvider>
```
jurajpiar, line 87: I don't understand why this needs to be service-wide context (what irony... :p). Isn't this only useful and used for creating an offer, as the [ticket](https://github.com/rsksmart/rif-marketplace-ui/issues/454) suggests?

jurajpiar (changes requested):
Could you please provide some clarification as to why it is necessary to use service-wide context to implement this feature?

src/components/organisms/storage/staking/Staking.tsx
```diff
@@ -116,6 +120,8 @@ const Staking: FC = () => {
   const [processingTx, setProcessingTx] = useState(false)
   const [txOperationDone, setTxOperationDone] = useState(false)
 
+  const isDisabled = !account || !isWhitelistedProvider
```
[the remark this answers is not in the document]
farinalvaro: Good idea 🙂

src/components/pages/storage/sell/StorageSellPage.tsx
```diff
@@ -173,6 +179,11 @@ const StorageSellPage = () => {
       <Typography gutterBottom color="secondary" variant="subtitle1">
         Fill out the fields below to list your storage service. All the information provided is meant to be true and correct.
       </Typography>
+      {
+        Boolean(account)
+        && isWhitelistedProvider === false
```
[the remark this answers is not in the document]
farinalvaro: Actually not, if we don't know if the provider is whitelisted (`isWhitelistedProvider 
 == undefined`) we disable some functionalities, but we don't want to show this message until we know if the account is not whitelisted.

src/components/pages/storage/StorageRoutes.tsx
farinalvaro: Yes, I should have clarified that we also need this for staking. So, I think that having this in a global Storage context has the benefit of calculating it only once, but I'm open to better ideas 🙂

jurajpiar (approved):
䷯✔️ LGTM

jurajpiar (approved):

PR #493, opened 2020-11-17
feat(storage): adds filter for own storage listing

jurajpiar:
Adds a filter option for account (to be filtered out).

Resolves https://rsklabs.atlassian.net/browse/RMKT-395

Works with cache PR https://github.com/rsksmart/rif-marketplace-cache/pull/399

### DIFF
--- a/src/context/Services/storage/offers/Context.tsx
+++ b/src/context/Services/storage/offers/Context.tsx
@@ -13,10 +13,11 @@ import { MinMaxFilter } from 'models/Filters'
 import { UIError } from 'models/UIMessage'
 import { AvgBillingPriceService } from 'api/rif-marketplace-cache/storage/avg-billing-plan-price'
 import createWithContext from 'context/storeUtils/createWithContext'
+import { Web3Store } from '@rsksmart/rif-ui'
 import {
   storageOffersActions, StorageOffersPayload,
   StorageOffersReducer, StorageOffersAction,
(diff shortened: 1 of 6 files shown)


src/context/Services/storage/offers/Context.tsx
```diff
@@ -113,13 +120,21 @@ export const Provider = ({ children }) => {
         //   payload: { refresh: true },
         // } as any)
         // setTimeout(() =>
+
+        dispatch({
+          type: 'FILTER',
+          payload: {
+            account,
+          },
+        })
```
farinalvaro, line 129: I might be missing something but why do we filter by the current account?
As we have the provider of every offer, could we just compare it to the current account?

farinalvaro: Also, I think we need to display that offer but just prevent the user from buying it

jurajpiar: > As we have the provider of every offer, could we just compare it to the current account?

Filtering via SQL is less expensive.

> Also, I think we need to display that offer but just prevent the user from buying it

Fair enough. I guess that's the way it is done in rns as well, after all. In that case, I'd do it locally of course.

jurajpiar:
Closing as there was a misunderstanding of requirements on my part and it's going to be simpler to create a new branch than modify this one.

PR #497, opened 2020-11-18
feat(storage): cleans up previous plans when creating offers & allows providers to see their inactive offers

### LINKED ISSUE #473: Remove billing plans when terminating an offer
Migrated from https://github.com/rsksmart/rif-marketplace-storage/issues/141
## Steps to reproduce
- As a provider, set an offer with a billing plan `A`
- Terminate that offer
- Create another offer with billing plan `B` (different period and/or currency than `A`)

## Expected behavior
- To have an offer with billing plan `B`

## Actual behavior
- The offer contains billing plan `A` and billing plan `B`

### LINKED ISSUE #477: Displaying current agreements after canceling offer
Once an offer is canceled we are not displaying the current Agreements anymore. But we should devise a way to be able to still see that offer if it has agreements. Until all agreements expired and all funds have been withdrawn by the Provider. Otherwise we dont give them the option to monitor existing agreements and withdraw funds.

Currently once an offer is Canceled it completely disappears from the UI. We should address this if it has Agreements.

farinalvaro:
- makes all filters optional so we can just filter by provider no matter size or price
- when creating offers, cleans up billing plans if the provider had a previous offer (inactive or not) - Closes #477 
- allows providers to see their inactive offers - Closes #473

### DIFF
--- a/src/components/pages/storage/myoffers/StorageMyOffersPage.tsx
+++ b/src/components/pages/storage/myoffers/StorageMyOffersPage.tsx
@@ -7,7 +7,6 @@ import Typography from '@material-ui/core/Typography'
 import { Web3Store } from '@rsksmart/rif-ui'
 import handProvidingFunds from 'assets/images/handProvidingFunds.svg'
 import CenteredPageTemplate from 'components/templates/CenteredPageTemplate'
-import withOffersContext, { StorageOffersContext, StorageOffersContextProps } from 'context/Services/storage/offers'
 import OffersList from 'components/organisms/storage/myoffers/OffersList'
 import AppContext, { AppContextProps, errorReporterFactory } from 'context/App/AppContext'
 import { StorageContract } from 'contracts/storage'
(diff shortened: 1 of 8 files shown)


src/components/pages/storage/myoffers/StorageMyOffersPage.tsx
```diff
@@ -59,16 +51,25 @@ const StorageMyOffersPage: FC = () => {
...
-        payload: { provider: account },
-      })
+    const getOwnOffers = async (): Promise<void> => {
+      const storageOffersService = appState?.apis?.['storage/v0/offers'] as StorageOffersService
+      storageOffersService.connect(errorReporterFactory(appDispatch))
+
+      const currentOwnOffers = await storageOffersService.fetch({
+        nonActive: true,
+        provider: account,
+      }) as StorageOffer[]
```
farinalvaro, line 65: I stopped using the context that fetches the offers here.
I think it's only useful for the Buy page now as there we don't consider inactive offers, we initialize limits that we don't need here (and also don't consider inactive offers for their calculation)
Might it be a good idea to initialize the `StorageOffersService` in the Storage "global" context and expose it?

nduchak (dismissed):
LGTM 👍

jurajpiar (approved):

julianmrodri (dismissed):
LGTM

jurajpiar (approved):

PR #498, opened 2020-11-18
fix: add contractAddress for upload service

julianmrodri:
* Fixes the upload service providing the additional parameter. The `storageManager` contract address.

### DIFF
--- a/src/context/Services/storage/upload/Context.tsx
+++ b/src/context/Services/storage/upload/Context.tsx
@@ -9,10 +9,10 @@ import Logger from 'utils/Logger'
 import { UIError } from 'models/UIMessage'
 import createWithContext from 'context/storeUtils/createWithContext'
 import { Web3Store } from '@rsksmart/rif-ui'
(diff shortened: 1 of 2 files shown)


src/context/Services/storage/upload/Context.tsx
```diff
@@ -4,6 +4,7 @@ import React, {
 } from 'react'
 import { createReducer } from 'context/storeUtils/reducer'
 import AppContext, { AppContextProps, errorReporterFactory } from 'context/App/AppContext'
+import networkConfig from 'config'
```
farinalvaro, line 7: We could directly import the storage contract address here

farinalvaro (commented):
Looks good. Just that small comment

src/context/Services/storage/upload/Context.tsx
julianmrodri: Sounds good.

farinalvaro (approved):
LGTM

PR #502, opened 2020-11-20
fix(storage): lets user follow buy w no acc

### LINKED ISSUE #446: Prevent buying your own Storage Offer
We should add a check similar to what we have in RNS to ensure that a Provider cannot **BUY** its own offer. Instead we should display `my offer` text like we do in RNS. 

For example in the screen below im logged in with the Provder account and I can click on BUY, which is incorrect.
image

### LINKED ISSUE #495: Enable selecting offers when not logged in and show `your offer`
* We should enable the users to SELECT offers if not logged in (similar to RNS).
* also we should display `your offer` instead of disabling the select button (similar RNS)

This is continuation of the previous issue to prevent selecting your own offer.

jurajpiar:
Users that didn't connect their wallet should be able to select a storage to purchase.

Resolves #495
Closes #446


itofarina (dismissed):
LGTM 🥇

julianmrodri (dismissed):
LGTM

julianmrodri (approved):
LGTM

PR #518, opened 2020-11-23
feat(storage): filters and fetches limits by available size

### LINKED ISSUE #460: Filter by Available Size not Total Capacity
Currently in the main **STORAGE-BUY** screen the filter is done by `totalCapacity`. But in reality what we are interested and what we see in the screen is the `Available Size`. we should change the filter so it filters by Available size.

farinalvaro:
Closes #460 
Depends on https://github.com/rsksmart/rif-marketplace-cache/pull/414

### DIFF
--- a/src/api/rif-marketplace-cache/storage/available-size.ts
+++ b/src/api/rif-marketplace-cache/storage/available-size.ts
@@ -17,8 +17,11 @@ export class AvailableCapacityService
   _fetch = (): Promise<[number, number]> => this.service.find()
 
   fetchSizeLimits = async (): Promise<MinMaxFilter> => {
-    // 1 arg is required by service interface but not used
-    const { min: minMB, max: maxMB } = await this.service.get(1)
+
+    const {
+      min: minMB,
+      max: maxMB,
+    } = await this.service.get(1) // 1 isn't used but required by service iface
     return {
       min: minMB / UNIT_PREFIX_POW2.KILO,
       max: maxMB / UNIT_PREFIX_POW2.KILO,
--- a/src/api/rif-marketplace-cache/storage/available-size.ts
+++ b/src/api/rif-marketplace-cache/storage/available-size.ts
@@ -17,7 +17,6 @@ export class AvailableCapacityService
   _fetch = (): Promise<[number, number]> => this.service.find()
 
   fetchSizeLimits = async (): Promise<MinMaxFilter> => {
-
     const {
       min: minMB,
       max: maxMB,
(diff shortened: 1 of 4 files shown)


src/context/Services/storage/offers/OffersContext.tsx
```diff
@@ -99,6 +101,10 @@ export const Provider = ({ children }) => {
   if (apiAvgBillingPrice && !apiAvgBillingPrice.service) {
     apiAvgBillingPrice.connect(errorReporterFactory(appDispatch))
   }
+
+  if (apiAvailableCapacity && !apiAvailableCapacity.service) {
```
jurajpiar, line 105: `apiAvailableCapacity` will always be true in compile time (`appState?.apis?.['storage/v0/availableCapacity'] as AvailableCapacityService`)
as well as runtime as the apis are all instantiated from within the `appContext`
Now, I know that's how it is being done in that file, but maybe you could remove the check for the api object in the other two if statements, as well?
This also means changing `appState?.apis?.[...` to `appState.apis[...`.

src/context/Services/storage/offers/OffersContext.tsx
```diff
@@ -99,6 +101,10 @@ export const Provider = ({ children }) => {
...
+    apiAvailableCapacity.connect(errorReporterFactory(appDispatch))
```
jurajpiar, line 106: Also here the errorReporterFactory could be called only once and reused (please change to other likes too)

src/api/rif-marketplace-cache/storage/available-size.ts
```diff
@@ -0,0 +1,27 @@
...
+
+  constructor() { super(client) }
+
+  _channel = 'availableCapacity' as StorageWSChannel
+
+  _fetch = (): Promise<[number, number]> => this.service.find()
+
+  fetchSizeLimits = async (): Promise<MinMaxFilter> => {
+    // this 1 arg is required by service interface, but it's not actually used
+    const { min: minMB, max: maxMB } = await this.service.get(1)
```
jurajpiar, line 21:
```suggestion
    const { min: minMB, max: maxMB } = await this.service.get(1) // 1 required by service iface
```

jurajpiar (changes requested):
Nice one. Could you just fix those bits where you copied old code stinks, please?

src/context/Services/storage/offers/OffersContext.tsx
farinalvaro: Done, thanks

src/context/Services/storage/offers/OffersContext.tsx
farinalvaro: Done, thanks

src/api/rif-marketplace-cache/storage/available-size.ts
farinalvaro: Done, thanks

jurajpiar (approved):

PR #520, opened 2020-11-23
feat(general): adds addr conversion to checksum

### LINKED ISSUE #503: Apply Checksum for addresses in Marketplace
We should apply correctly the checksum and display addresses in the correct Checksum according to the network.

jurajpiar:
Creates a function that wraps both toChecksum and shortenAddress functions. Then uses this across the project, where an address is displayed to the user.
Also changes AddressItem into RifAddress component, which also uses this function.

Resolves #503
Resolves [RMKT-346](https://rsklabs.atlassian.net/browse/RMKT-346)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1269,9 +1269,9 @@
       }
     },
     "@babel/traverse": {
-      "version": "7.12.8",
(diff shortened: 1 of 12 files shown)


itofarina (dismissed):
LGTM

julianmrodri (approved):
LGTM

PR #522, opened 2020-11-24
feat(storage): gets currency from plan transport

### LINKED ISSUE #501: Billing periods shows invalid on agreement create page
On agreement create step we had a select with billing plans, this select show incorrect data for billing plans with different tokens(for exm: when we had two plans one for RBTC and another for RIF, then it shows 2 plans with RBTC)

jurajpiar:
Adds the retrieval of the currency associated to a plan. The rest of the app has been designed anticipating this change, thus no other action needed.

~There is an issue with the rates for other currencies not showing, ie they show as 0, even thought the context (and indeed the component) observes the new rates. This however is not in the scope of this PR and will be handled as a separate issue (https://github.com/rsksmart/rif-marketplace-ui/issues/523).~

Resolves [RMKT-425](https://rsklabs.atlassian.net/browse/RMKT-425)
Resolves #501

### DIFF
--- a/src/api/models/storage/transports.ts
+++ b/src/api/models/storage/transports.ts
@@ -2,9 +2,11 @@ export interface BillingPlanTransport {
   id: number
   period: string
   price: string
+  rateId: string
   offerId: string
   createdAt: string
   updatedAt: string
+  tokenAddress: string
 }
 
 export interface OfferTransport {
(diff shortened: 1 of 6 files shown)


itofarina (dismissed):
LGTM

julianmrodri (dismissed):
LGTM

julianmrodri (dismissed):
LGTM

julianmrodri (approved):
LGTM

julianmrodri (changes requested):
This is not working as expected. Even though you see the Correct Currency and can switch among them. then it ALWAYS uses RBTC as the payment. We need to double check we are sending the correct information executing the correct Singleton contract with the correct values.

From what I see looks like we are not passing the `token` attribute as part of the TxOptions when calling `newAgreement`. We should pass there either `rif` or `rbtc`.

julianmrodri:
Lets metge it as-is and we can take care of the rest in the Multicurrency fix.

julianmrodri (dismissed):
LGTM

julianmrodri (approved):
LGTM! Nice!

PR #524, opened 2020-11-24
feat(storage-sell): pre-fills peerId from URL

jurajpiar:
What the tin says.

Resolves [RMKT-402](https://rsklabs.atlassian.net/browse/RMKT-402)

Moved the cidValidation into its own utils file to prevent the CID TextDecoder issue when testing.


jurajpiar:
no idea why it starts failing on that CID sh#t again.

julianmrodri:
What would be the URL for the `Pinner` service? So we can include that there?
Currently we have:  `uiUrl: http://localhost:3000/offer?peerId=<<peerId>>`

julianmrodri (approved):
LGTM

Issue #377, opened 2020-09-29
Optimize spent gas when editing storage offers


farinalvaro:
At the moment, when the user edits an offer, we are calling the `setOffer` method that consumes more gas than `setCapacity` for example. So, if the user has only changed the capacity, it would be better to only call `setCapacity`

PR #525, opened 2020-11-24
feat(storage): uses totalCapacity for offer edition/creation

### LINKED ISSUE #508: Use `totalCapacity` when editing offer as Provider
When we go to `My-Offers` and then select **EDIT**.  We are displaying the available size in the Offer. But in this screen as provider we always want to see the `totalCapacity` listed and modify the `totalCapacity` value. This will require to also add a validation when submitting the TX that the new `totalCapacity` has to be greater than the allocated agreements.

image

farinalvaro:
Closes #508

### DIFF
--- a/src/components/organisms/storage/sell/EditOfferStepper.tsx
+++ b/src/components/organisms/storage/sell/EditOfferStepper.tsx
@@ -38,9 +38,10 @@ const EditOfferStepper: FC<EditOfferStepperProps> = ({ endHandler }) => {
   const handleNext = (): void => setActiveStep(1)
   const handleBack = (): void => setActiveStep(0)
 
-  const minCapacity = Number(originalOffer?.utilizedCapacityGB) || 0
-  const nextIsDisabled = !totalCapacity
-    || Number(totalCapacity) <= minCapacity
+  const minCapacity = originalOffer?.utilizedCapacityGB.toNumber() || 0
+  const totalCapacityNumber = totalCapacity.toNumber()
+  const nextIsDisabled = !totalCapacityNumber
+    || totalCapacityNumber <= minCapacity
     || !system
     || !peerId
 
--- a/src/components/organisms/storage/sell/EditOfferStepper.tsx
+++ b/src/components/organisms/storage/sell/EditOfferStepper.tsx
@@ -30,14 +30,17 @@ const EditOfferStepper: FC<EditOfferStepperProps> = ({ endHandler }) => {
   const classes = useStyles()
   const [activeStep, setActiveStep] = useState(0)
   const {
-    state: { availableSize, system, peerId },
+    state: {
+      totalCapacity, system, peerId, originalOffer,
+    },
   } = useContext<OfferEditContextProps>(OfferEditContext)
 
   const handleNext = (): void => setActiveStep(1)
   const handleBack = (): void => setActiveStep(0)
 
-  const nextIsDisabled = !availableSize
-    || availableSize <= 0
+  const minCapacity = Number(originalOffer?.utilizedCapacityGB) || 0
+  const nextIsDisabled = !totalCapacity
+    || Number(totalCapacity) <= minCapacity
     || !system
     || !peerId
 
--- a/src/components/organisms/storage/sell/GeneralFeatures.tsx
+++ b/src/components/organisms/storage/sell/GeneralFeatures.tsx
@@ -1,3 +1,4 @@
+import Big from 'big.js'
 import React, { FC, useContext, useState } from 'react'
 import InfoIcon from '@material-ui/icons/Info'
 import Grid from '@material-ui/core/Grid'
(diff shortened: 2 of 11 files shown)


src/components/organisms/storage/sell/EditOfferStepper.tsx
```diff
@@ -30,14 +30,17 @@ const EditOfferStepper: FC<EditOfferStepperProps> = ({ endHandler }) => {
   const classes = useStyles()
   const [activeStep, setActiveStep] = useState(0)
   const {
-    state: { availableSize, system, peerId },
+    state: {
+      totalCapacity, system, peerId, originalOffer,
```
jurajpiar, line 34: perhaps just pass the `utilizedCapacityGB`, since the component doesn't use anything else from the object.

src/components/organisms/storage/sell/GeneralFeatures.tsx
```diff
@@ -12,25 +13,26 @@ import {
...
 import OfferEditContext from 'context/Market/storage/OfferEditContext'
 import { OfferEditContextProps } from 'context/Market/storage/interfaces'
-import { SetAvailableSizePayload, SetPeerIdPayload } from 'context/Market/storage/offerEditActions'
+import { SetTotalCapacityPayload, SetPeerIdPayload } from 'context/Market/storage/offerEditActions'
 
 const GeneralFeatures: FC = () => {
   const {
     state: {
-      availableSize, system, peerId, originalOffer,
+      totalCapacity, system, peerId, originalOffer,
```
jurajpiar, line 21: same here. just pass `utilisedCapacityGB`

src/components/pages/storage/myoffers/StorageEditOfferPage.tsx
```diff
@@ -33,7 +33,7 @@ const StorageEditOfferPage: FC<{}> = () => {
   const reportError = useCallback((e: UIError) => errorReporterFactory(appDispatch)(e), [appDispatch])
   const {
     state: {
-      originalOffer, billingPlans, availableSize, peerId,
+      originalOffer, billingPlans, totalCapacity, peerId,
```
jurajpiar, line 36: again no need for the whole original offer object. the `transformOfferDataForContract` only uses the `subscriptionOptions`

src/components/pages/storage/myoffers/StorageEditOfferPage.tsx
```diff
@@ -113,7 +113,7 @@ const StorageEditOfferPage: FC<{}> = () => {
     }
   }, [isPendingConfirm, history, isProcessing])
 
-  const isSubmitEnabled = !!(billingPlans.length && availableSize)
+  const isSubmitEnabled = !!(billingPlans.length && totalCapacity)
```
jurajpiar, line 116:
```suggestion
  const isSubmitEnabled = Boolean(billingPlans.length && totalCapacity)
```

src/contracts/storage/utils.ts
```diff
@@ -53,11 +53,11 @@ export function encodeHash(hash: string): string[] {
 }
 
 export const transformOfferDataForContract = (
-  availableSizeGB: number,
+  totalCapacityGB: Big,
   billingPlans: StorageBillingPlan[],
   originalOffer?: StorageOffer,
```
jurajpiar, line 58: just pass `subscriptionOptions`

jurajpiar (changes requested):
👾 Just noticed the unnecessary object we're passing around.

src/components/organisms/storage/sell/EditOfferStepper.tsx
farinalvaro: The problem with that is that `originalOffer` could be `undefined` because the same component is being used for offer creations and in that case we don't store a copy of the `originalOffer`

src/components/organisms/storage/sell/GeneralFeatures.tsx
farinalvaro: Same thing + we need to know if `originalOffer` is not undefined and in that case disable the `peerId` field.

src/components/pages/storage/myoffers/StorageEditOfferPage.tsx
farinalvaro: We need the `originalOffer` that was set previously on `My Offers` page, if it was not set then we redirect to My Offers page, similar to what we do with orders in checkout

src/contracts/storage/utils.ts
farinalvaro: Done, thanks.
Just to left some guidance for the future I want to link https://github.com/rsksmart/rif-marketplace-ui/issues/377 with this change

jurajpiar (dismissed):

jurajpiar (approved):

PR #544, opened 2020-11-26
fix(storage-purchases): removes id as address

farinalvaro:
Removes conversion of agreement id to RifAddress and so checksum addredss


jurajpiar:
Closing as this is demonstrably not where the problem lies. sry

PR #551, opened 2020-11-27
fix: removes precision from agreements available funds

farinalvaro:
When combining RIF token and RBTC we had some issues displaying the amounts with `toPrecision(X)` so just showing the full amount until we design proper solution

**Before**
Screen Shot 2020-11-27 at 13 22 46

**Now**
Screen Shot 2020-11-27 at 13 23 16


julianmrodri (approved):
LGTM

jurajpiar (approved):

PR #553, opened 2020-11-30
fix(notifications): removes duplicated auth call

jurajpiar:
Partially resolves [RMKT-437](https://rsklabs.atlassian.net/browse/RMKT-437)

Partially, because the context still seems to be created twice (thought it remains unverified whether this is a problem).


itofarina (approved):

julianmrodri (approved):
LGTM

PR #557, opened 2020-11-30
fix(storage): ensures availableCapacity >= 0

jurajpiar:
Maps received (from cache) availableSize to a number >= 0.

Resolves [RMKT-436](https://rsklabs.atlassian.net/browse/RMKT-436)

🙈


itofarina (commented):
The fix looks good but may we remove the pagination?

jurajpiar:
> The fix looks good but may we remove the pagination?

¿Por que?

itofarina (approved):
I just thought it would be better to add it on a separated PR with the full functionality, like moving between pages for example. 
But, you're right, it doesn't harm the solution so LGTM, up to you if you want to keep it :)

PR #564, opened 2020-12-02
fix(storage-buy): resets periods and selected plan on currency change

### LINKED ISSUE #528: Setting Periods Prepay to 0 when changing Currency
When you buy an agreement and change the Currency. It sets the Period to 0. We should initialize at least to 1 like the default of the page or maintain always the existing value. Anyway the user has the expiration date to confirm.

image

farinalvaro:
Closes #528


jurajpiar (approved):
Consider following change:

julianmrodri (approved):
LGTM

jurajpiar (approved):

PR #581, opened 2020-12-07
feat: checks available funds before sending tx

### LINKED ISSUE #417: Check for available funds before sending Staking transaction
This would prevent gas loss

### LINKED ISSUE #530: Check available funds when creating Agreement
When creating and agreement and selecting a plan we should notify the user before sending the TX that they do not have enough funds. This should consider the multicurrency feature and work for the selected currency/token. This is in the Agreement creation process.

farinalvaro:
Migrated from #569 
Closes #530 
Closes #417 

Adds `CustomError` class to handle custom errors with user friendly messages

### DIFF
--- a/src/contracts/storage/Staking.ts
+++ b/src/contracts/storage/Staking.ts
@@ -9,6 +9,9 @@ import { stakingAddress, storageSupportedTokens } from 'contracts/config'
 import { SUPPORTED_TOKENS, TxOptions } from 'contracts/interfaces'
 import { getTokens } from 'utils/tokenUtils'
 import ContractWithTokens from 'contracts/wrappers/contract-using-tokens'
+import Big from 'big.js'
+import CustomError from 'models/CustomError'
+import { validateBalance } from 'contracts/utils/accountBalance'
 
 export type StorageStakingContractErrorId = 'contract-storage-staking'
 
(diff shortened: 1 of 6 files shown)


src/contracts/storage/Staking.ts
```diff
@@ -32,17 +35,22 @@ class StakingContract extends ContractWithTokens {
...
   private static instance: StakingContract
 
-  public stake(
+  public async stake(
     amount: string | number,
     txOptions: TxOptions,
   ): Promise<TransactionReceipt> {
-    if (amount < 0) {
-      throw new Error('amount should greater then 0')
+    if (Big(amount).lte(Big(0))) {
```
jurajpiar, line 42: I think you just want `.lt` not `.lte`.

src/contracts/storage/Staking.ts
```diff
@@ -32,17 +35,22 @@ class StakingContract extends ContractWithTokens {
...
     }
 
-    const { tokenAddress } = this.getToken(txOptions.token)
+    const { from: account, token } = txOptions
+    const { tokenAddress } = this.getToken(token)
     const amountWei = convertToWeiString(amount)
 
+    await validateBalance({
+      web3: this.web3, minAmountWei: amountWei, account, token,
+    })
```
jurajpiar, line 52: Feel free to ignore.
```suggestion
    const minAmountWei = convertToWeiString(amount)
    const { web3 } = this

    await validateBalance({
      web3, minAmountWei, account, token,
    })
```

src/contracts/utils/accountBalance.ts
```diff
@@ -20,4 +22,26 @@ const getBalance = (
...
+
+  const balance = await getBalance(web3, account, token)
+
+  if (Big(balance).lt(minAmountWei)) throw new CustomError('Not enough funds.')
+}
+
+export default {
+  getBalance,
+  validateBalance,
+}
```
jurajpiar, line 47: maybe just do one or the other but not both exports?

jurajpiar (changes requested):
Nice one

src/contracts/storage/Staking.ts
farinalvaro: Hmmm I don't think so, with `lt` we would allow to stake `0`, right?

src/contracts/storage/Staking.ts
farinalvaro: Sorry for ignoring 😅 I don't like much the `const minAmountWei = convertToWeiString(amount)` as it would result in `this.methods.stake(minAmountWei, ...`

jurajpiar (approved):

PR #587, opened 2020-12-09
fix(contract-tokens): lowercases from address before sending tx

farinalvaro:
Another option (perhaps cleaner) could be to make [TransactionOptions](https://github.com/rsksmart/rif-marketplace-ui/blob/master/src/contracts/interfaces.ts#L13) a class and lowercase the from attribute on the constructor


farinalvaro:
Closing as it's being resolved by providing the logged account already in lowercase from the Web3Provider in rif-ui library

PR #591, opened 2020-12-10
fix(storage-agreements): includes token on txOptions when renewing

### LINKED ISSUE #585: Error renewing funds for Token (RIF) Agreement
There is a problem when trying to renew funds for an Agreement in tokens (for example RIF). We get the error that the `Agreement does not exist`.  It works fine for RBTC. We dont event get to the approval Transaction.

image

image

farinalvaro:
- Includes token on `txOptions` when renewing agreements - Closes #585 
- Updates `rif-ui` dependency to a new version where the account is given in lowercase by the `Web3Provider`

~Depends on https://github.com/rsksmart/rif-ui/pull/206 that would release new npm version. 
Do not merge until that dependency is updated properly~

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2769,8 +2769,9 @@
       }
     },
     "@rsksmart/rif-ui": {
-      "version": "0.5.0",
-      "resolved": "github:rsksmart/rif-ui#26dc6f45f899ab54f3c1938c317505598e5108c2"
+      "version": "0.5.1",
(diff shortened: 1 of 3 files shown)


julianmrodri:
LGTM!! We should display the Logged account in the top right corner using Checksum. Now shows up with Lower.
image

jurajpiar (approved):

PR #594, opened 2020-12-11
feat(confirmations): adds confirmations context

### LINKED ISSUE #586: Spinner improvements for waiting confirmations
We need to include an experience with Spinner similar to what we have in Staking but to other transactions.

- When creating Agreement, to see as soon as you get confirmation that there it will soon show up (spinner in My Purchases)
- The same in updates of Renewal and Withdraw (maybe we can do the spinner in the specific fields being updated)

farinalvaro:
`BlockchainContext` and `ConfirmationsContext` should be the same, but for the matter of transition I've created `ConfirmationsContext` to handle confirmations in a new way without interfering with `BlockchainContext` that's being used in many other services. 

Will close #586 

Will add https://github.com/rsksmart/rif-marketplace-ui/issues/607

### DIFF
--- a/src/components/pages/storage/myPurchases/Page.tsx
+++ b/src/components/pages/storage/myPurchases/Page.tsx
@@ -31,6 +31,7 @@ import withWithdrawContext, { StorageWithdrawContext, StorageWithdrawContextProp
 import GridRow from 'components/atoms/GridRow'
 import { ConfirmationsContext, ConfirmationsContextProps } from 'context/Confirmations'
 import getConfirmationsFor from 'context/Confirmations/utils'
+import InfoBar from 'components/molecules/InfoBar'
 
 const useTitleStyles = makeStyles(() => ({
   root: {
@@ -179,10 +180,12 @@ const MyStoragePurchases: FC = () => {
       {
         Boolean(newAgreementsConfsCount)
         && (
-          <Typography>
-            {`Waiting confirmations for
-             ${newAgreementsConfsCount} new agreement(s)`}
-          </Typography>
+          <InfoBar
+            isVisible
+            type="info"
+            text={`Waiting confirmations for
+            ${newAgreementsConfsCount} new agreement(s)`}
+          />
         )
       }
       <RoundedCard color="secondary">
--- a/src/components/pages/storage/myPurchases/Page.tsx
+++ b/src/components/pages/storage/myPurchases/Page.tsx
@@ -29,6 +29,8 @@ import WithLoginCard from 'components/hoc/WithLoginCard'
 import ProgressOverlay from 'components/templates/ProgressOverlay'
 import withWithdrawContext, { StorageWithdrawContext, StorageWithdrawContextProps } from 'context/storage/mypurchases/withdraw'
 import GridRow from 'components/atoms/GridRow'
+import { ConfirmationsContext, ConfirmationsContextProps } from 'context/Confirmations'
+import getConfirmationsFor from 'context/Confirmations/utils'
 
 const useTitleStyles = makeStyles(() => ({
   root: {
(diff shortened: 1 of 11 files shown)


src/components/pages/storage/myPurchases/Page.tsx
```diff
@@ -164,6 +171,16 @@ const MyStoragePurchases: FC = () => {
 
   return (
     <CenteredPageTemplate>
+      {
+        !!awaitingServices
+        && awaitingServices.Agreements
+        && !!awaitingServices.Agreements.length
```
jurajpiar, line 177: `Boolean(awaitingServices?.Agreements?.length)`

src/context/Confirmations/Context.tsx
```diff
@@ -0,0 +1,107 @@
...
+  const [isInitialised, setIsInitialised] = useState(false)
+
+  // Get service connection
+  if (!service) {
+    connect(errorReporter)
+  }
+
+  // Initialise
+  useEffect(() => {
+    if (service && !isInitialised && account) {
```
jurajpiar, line 63: the `isInitialised` will prevent re-authentication upon account change. might not be an issue atm, as when account changes we go back to home page. Once we fix that (assuming we will) though, it will be unable to re-auth as the `isInitialised` will be true by that point.

src/context/Confirmations/actions.ts
```diff
@@ -0,0 +1,48 @@
...
+  }: NewConfirmationPayload) => {
+    const { confirmations } = state
+    const confirmationRecord = confirmations[transactionHash]
+
+    if (!confirmationRecord) { // this conf is not being tracked
+      return { ...state }
+    }
+
+    if (confirmationsCount >= targetConfirmation) {
+      delete confirmations[transactionHash]
```
jurajpiar, line 17: Could we do this without mutating the argument, please.

src/context/Confirmations/actions.ts
```diff
@@ -0,0 +1,48 @@
...
+    } else {
+      confirmationRecord.currentCount = confirmationsCount
+      confirmationRecord.targetCount = targetConfirmation
```
jurajpiar, line 20: same here

jurajpiar (changes requested):
Good job. Please see my comments:

src/components/pages/storage/myPurchases/Page.tsx
```diff
@@ -164,6 +181,18 @@ const MyStoragePurchases: FC = () => {
...
     <CenteredPageTemplate>
+      {
+        Boolean(confAwaitersKeys.length)
+        && (
+          <InfoBar
+            isVisible
+            type="info"
+            text={`Awaiting confirmations for
+          ${confAwaitersKeys.length} new
+           ${maybePluralize(confAwaitersKeys.length, 'agreement')}`}
```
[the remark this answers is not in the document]
farinalvaro: I understand your point but I would prefer to not lose precision for a problem that doesn't exist now. Also the suffix is parametrized so we can set any depending on the noun or the language we want to use.
Is that convincing or should I make that change anyway?

src/context/Confirmations/Context.tsx
farinalvaro: Good point. May be that's something that needs to be considered if and when we want to change the functionality of refreshing and going to home on account change, as this logic is being used in many contexts with services, right?

src/context/Confirmations/actions.ts
farinalvaro: Done. confirmationRecord comes from the copy now

src/context/Confirmations/utils.ts
```diff
@@ -0,0 +1,16 @@
...
+  contractAction: ContractAction,
+  confirmations: ConfirmationsRecord,
+): ConfirmationsRecord => Object.keys(confirmations).reduce(
+  (acc: ConfirmationsRecord, curr: TxHash) => {
+    if (confirmations[curr].contractAction === contractAction) {
+      acc[curr] = confirmations[curr]
+    }
+    return acc
+  }, {},
+)
```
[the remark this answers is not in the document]
farinalvaro: Good point, thanks 🙂

jurajpiar (approved):
✌️

julianmrodri (approved):
LGTM

PR #605, opened 2020-12-16
feat(storage): optimize edit offer call

nduchak:
Closes https://github.com/rsksmart/rif-marketplace-ui/issues/377

### DIFF
--- a/src/components/pages/storage/myoffers/StorageEditOfferPage.tsx
+++ b/src/components/pages/storage/myoffers/StorageEditOfferPage.tsx
@@ -18,7 +18,7 @@ import { useHistory } from 'react-router-dom'
 import ROUTES from 'routes'
 import Logger from 'utils/Logger'
 import TransactionInProgressPanel from 'components/organisms/TransactionInProgressPanel'
-import { transformOfferDataForContract } from 'contracts/storage/utils'
+import { isBillingPlansChanges, transformOfferDataForContract } from 'contracts/storage/utils'
 import { SupportedTokens } from 'contracts/interfaces'
 import { StorageOffer } from 'models/marketItems/StorageItem'
 import Web3 from 'web3'
@@ -46,19 +46,6 @@ const StorageEditOfferPage: FC<{}> = () => {
 
   const [isPendingConfirm, setIsPendingConfirm] = useState(false)
   const [isProcessing, setIsProcessing] = useState(false)
-
-  const isBillingPlansChanges = (current: StorageBillingPlan[], previous: StorageBillingPlan[]): boolean => {
-    if (current.length !== previous.length) {
-      return true
-    }
-    return previous.some(
-      (plan) => !current.find(
-        (p) => p.currency === plan.currency
-              && p.period === plan.period
-              && p.price.toString() === plan.price.toString(),
-      ),
-    )
-  }
 
   const makeUpdateOfferTx = () => {
     const storageContract = StorageContract.getInstance(web3 as Web3)
@@ -123,13 +110,13 @@ const StorageEditOfferPage: FC<{}> = () => {
 
       setIsProcessing(true)
 
-      const setOfferReceipt = await makeUpdateOfferTx()
-      logger.info('setOffer receipt: ', setOfferReceipt)
+      const updateOfferReceipt = await makeUpdateOfferTx()
+      logger.info('setOffer receipt: ', updateOfferReceipt)
 
       bcDispatch({
         type: 'SET_TX_HASH',
         payload: {
-          txHash: setOfferReceipt.transactionHash,
+          txHash: updateOfferReceipt.transactionHash,
         } as AddTxPayload,
       })
       setIsPendingConfirm(true)
--- a/src/components/pages/storage/myoffers/StorageEditOfferPage.tsx
+++ b/src/components/pages/storage/myoffers/StorageEditOfferPage.tsx
@@ -48,6 +48,16 @@ const StorageEditOfferPage: FC<{}> = () => {
   const [isPendingConfirm, setIsPendingConfirm] = useState(false)
   const [isProcessing, setIsProcessing] = useState(false)
 
+  const isPlanChange = isBillingPlansChanged(
+    billingPlans,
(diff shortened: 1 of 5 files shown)


src/components/pages/storage/myoffers/StorageEditOfferPage.tsx
```diff
@@ -59,25 +122,8 @@ const StorageEditOfferPage: FC<{}> = () => {
...
 
-      const setOfferReceipt = await storageContract.setOffer(
-        totalCapacityMB,
-        periods,
-        prices,
-        tokens as SupportedTokens[],
-        peerId,
-        { from: account },
-      )
+      const setOfferReceipt = await makeUpdateOfferTx()
```
farinalvaro, line 126:
```suggestion
      const updateOfferReceipt = await makeUpdateOfferTx()
```

src/components/pages/storage/myoffers/StorageEditOfferPage.tsx
```diff
@@ -45,6 +47,67 @@ const StorageEditOfferPage: FC<{}> = () => {
...
+      return true
+    }
+    return previous.some(
+      (plan) => !current.find(
+        (p) => p.currency === plan.currency
+              && p.period === plan.period
+              && p.price.toString() === plan.price.toString(),
+      ),
+    )
+  }
```
farinalvaro, line 61: Love it! I would suggest to extract this function to `contracts/storage/utils`

src/components/pages/storage/myoffers/StorageEditOfferPage.tsx
```diff
@@ -45,6 +47,67 @@ const StorageEditOfferPage: FC<{}> = () => {
...
+      billingPlans,
+      subscriptionOptions,
+    )
+
+    const isPlanChange = isBillingPlansChanges(
+      billingPlans,
+      originalOffer
+          ?.subscriptionOptions
+          .map(
+              (p) => ({ ...p, price: p.price.mul(UNIT_PREFIX_POW2.KILO) })
```
farinalvaro, line 80: I'm thinking if instead of this `map` it could be better to do this during the comparison `(isBillingPlansChanges)` to avoid iterating over the list twice. What do you think?

src/components/pages/storage/myoffers/StorageEditOfferPage.tsx
```diff
@@ -45,6 +47,67 @@ const StorageEditOfferPage: FC<{}> = () => {
...
+    return previous.some(
+      (plan) => !current.find(
+        (p) => p.currency === plan.currency
+              && p.period === plan.period
+              && p.price.toString() === plan.price.toString(),
+      ),
+    )
+  }
+
+  const makeUpdateOfferTx = () => {
```
farinalvaro, line 63: I think this function shouldn't be in a page component, could it be better to move to `contracts/storage/Storage.ts`?

farinalvaro (commented):
Nice job 👏!
I just left a couple of suggestions, happy to have a chat about them 🙂

src/components/pages/storage/myoffers/StorageEditOfferPage.tsx
nduchak: Done

src/components/pages/storage/myoffers/StorageEditOfferPage.tsx
nduchak: Done

src/components/pages/storage/myoffers/StorageEditOfferPage.tsx
nduchak: Hmm... `isBillingPlansChanges` expect two arrays of BillingPlan interface(where prices in the same unit), and I won't put some additional transformation logic there because then it will be not reusable as it depends on some specific logic.
Imagine that you want to reuse this method in some other place then you should always remember that the first array contains prices in MB (which is strange as these two arrays have the same interface)

src/components/pages/storage/myoffers/StorageEditOfferPage.tsx
nduchak: The contract wrapper has only one responsibility to create and send a transaction for the specific contract. If we put this logic there we make it dependent on some logic from outside which is not SOLID. I prefer to keep this method in some place related to this component(not sure where maybe create some utils specific for this component)

src/components/pages/storage/myoffers/StorageEditOfferPage.tsx
nduchak: The thing is that we should use the same unit of price for billing plan coming from this component and for billing plan coming from the context

src/components/pages/storage/myoffers/StorageEditOfferPage.tsx
```diff
@@ -45,6 +49,54 @@ const StorageEditOfferPage: FC<{}> = () => {
   const [isPendingConfirm, setIsPendingConfirm] = useState(false)
   const [isProcessing, setIsProcessing] = useState(false)
 
+  const isPlanChange = isBillingPlansChanged(
+    billingPlans,
+    originalOffer
+        ?.subscriptionOptions
+        .map(
+            (p) => ({ ...p, price: p.price.mul(UNIT_PREFIX_POW2.KILO) })
+        ) as StorageBillingPlan[],
```
farinalvaro, line 58: Now with the change in the reducer this should already come in MB, right?

src/components/pages/storage/myoffers/StorageEditOfferPage.tsx
farinalvaro: Solved in https://github.com/rsksmart/rif-marketplace-ui/pull/605/commits/ff2a6277c8b8fc15ec6ecf272eab2065af9f4b9a thanks!

farinalvaro (dismissed):

julianmrodri:
I will test this to ensure its working.

julianmrodri (approved):
LGTM

jurajpiar (approved):

PR #618, opened 2020-12-23
fix(rns): stopps checksum of domain names

### LINKED ISSUE #611: Error when trying to buy RNS domain
When trying to buy an RNS domain after clicking buy we get an error because it tries to convert to `checksum` the RNS Domain. We need to fix this so it works as before allowing to buy/sell.

Steps to reproduce:
1 - Go to Name Services - Buy
2- Select a domain available for sale
3- You will get empty screen and the error.

image

jurajpiar:
Removes checksum from known domain names.

Resolves #611


julianmrodri (approved):
LGTM

PR #627, opened 2020-12-30
refactor(storage): avoids editOfferContext use in children comp

### LINKED ISSUE #542: Edit offer context refactor
We should read from OfferEditContext at the parent component and then try to pass the child props instead of reading from the context.
This will help with the re-usability of the children components

farinalvaro:
Closes #542


src/components/organisms/storage/sell/EditableBillingPlan.tsx
```diff
@@ -30,10 +29,9 @@ const EditableBillingPlan: FC<EditableBillingPlanProps> = ({
 }) => {
   const {
     state: { allBillingPeriods, usedPeriodsPerCurrency },
-    dispatch,
   } = useContext(OfferEditContext)
```
farinalvaro, line 32: This is the only child component where we are reading from context now. These props are used to track the available periods per currency and handle that logic outside of the components so I'm not sure about the benefit of removing it from here and passing the props all the way down.

julianmrodri (approved):
LGTM

jurajpiar (approved):
LGTM

PR #628, opened 2021-01-03
feat(storage): ensures positive prepaid period

jurajpiar:
ensures positive prepaid period when buying storage

Fixes [RMKT-476](https://rsklabs.atlassian.net/browse/RMKT-476)


julianmrodri (approved):
LGTM

PR #644, opened 2021-01-11
chore(deps): replaces rsk-util dependency with @rsksmart/rsk-utils

### LINKED ISSUE #636: Checksum in Staging not shown correctly
In Staging it should use the same checksum as in `testnet`. 

However if logged with account: `0x90f8Bf6a479F320eaD074411a4B0e7944Ea8C9c1` we can see in the top right it is not using the correct Checksum (compare it to the first row).

I believe this is also causing that it does not detect that its your own offer, which should be displayed instad of the button in the first row.

image

farinalvaro:
- Replaces `rskjs-util`(deprecated) dependency with `@rsksmart/rsk-utils`
- Bumps `rif-ui` dependency to `v0.5.4` - Closes #636


julianmrodri (approved):
LGTM

jurajpiar (approved):

PR #647, opened 2021-01-13
feat(confirmations): adds transactions panel popover

farinalvaro:
- adds txs icon with badge in header
- adds txs panel
- extracts logic and styles from notifications popover to be reused


julianmrodri (commented):
Looking really good. Lets do the changes we discussed offline:
  Font size reduce to make it consistent
  Align the space between Bell and Transaction icon so they are not too separate (add button to bell icon).

jurajpiar (approved):
LGTM

PR #660, opened 2021-01-18
feat(storage-upload): reads file size limit from service

farinalvaro:
Now on https://github.com/rsksmart/rif-marketplace-ui/pull/662

~- Gets file size limit from upload-service~
~- Shows max size text~
~- Disables upload button until there is content to be uploaded~

~By default it's set to 1GB in UI. This allows us to always handle a value without having to "pause" the flow until we have the max size defined~

### DIFF
--- a/src/context/Services/storage/upload/Context.tsx
+++ b/src/context/Services/storage/upload/Context.tsx
@@ -30,6 +30,9 @@ export const initialState: State = {
   contextID,
   status: {},
   fileSizeLimit: Big(UNIT_PREFIX_POW2.GIGA), // 1GB by default
+  isLoading: {
+    sizeLimit: false,
+  },
 }
 
 const initialAsyncActions: AsyncActions = {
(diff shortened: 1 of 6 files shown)


src/context/Services/storage/upload/Context.tsx
```diff
@@ -23,6 +29,10 @@ export type ContextName = typeof contextID
 export const initialState: State = {
   contextID,
   status: {},
+  fileSizeLimit: Big(UNIT_PREFIX_POW2.GIGA), // 1GB by default
+  isLoading: {
+    sizeLimit: false,
+  },
```
farinalvaro, line 35: Created an object for this so we can add a flag for uploading files action

julianmrodri (approved):
LGTM! 🚀

src/context/Services/storage/upload/Context.tsx
farinalvaro: Actually I noticed that this should be handled at component level and not in context so ignore my comment 😄

farinalvaro:
Closing in favor of https://github.com/rsksmart/rif-marketplace-ui/pull/662

PR #662, opened 2021-01-19
feat(storage-upload): adds spinner & disables btn on loading file

farinalvaro:
- Gets file size limit from upload-service
- Shows max size text
- Disables upload button until there is content to be uploaded
- Adds spinner on uploader and disables button until files are loaded

By default it's set to 1GB in UI. This allows us to always handle a value without having to "pause" the flow until we have the max size defined


jurajpiar (approved):
LGTM

Issue #607, opened 2020-12-17
Upgrade the old confirmations mechanism.


jurajpiar:
https://github.com/rsksmart/rif-marketplace-ui/pull/594 will add a new, more robust confirmations mechanism, without touching the old one.
We need to addresses the migration to the new mechanism across the repo ASAP.

PR #691, opened 2021-02-01
feat(rns): tracks cancel placed domain confirmations

farinalvaro:
~Rebased on #686~
Related to #607 

Updates awaiting confirmations mechanism for canceling domain placement in RNS


jurajpiar (approved):
LGTM just make sure the change in the CLEAR_ORDER useEffect doesn't affect the rest of the app.

PR #697, opened 2021-02-04
fix(rns): adds limit to offers request

### LINKED ISSUE #528: Re-enable pagination for RNS offers

### LINKED ISSUE #664: Filter by Price in RNS should use USD value
Now that we support multicurrency we should be using USD price as the filter. We cannot use simply RIF amount.

image

jurajpiar:
to restore pagination in domain offers

Resolves https://github.com/rsksmart/rif-marketplace-cache/issues/528
Related to https://github.com/rsksmart/rif-marketplace-cache/pull/529

### DIFF
--- a/src/api/rif-marketplace-cache/rns/offers.ts
+++ b/src/api/rif-marketplace-cache/rns/offers.ts
@@ -41,6 +41,7 @@ enum LimitType {
   min = 1,
   max = -1,
 }
+const PAGE_LIMIT = 10
 
 const fetchPriceLimit = async (
   service,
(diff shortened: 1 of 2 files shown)


julianmrodri (changes requested):
Initial pagination works fine, but the Pagination is not updated once you change the filter. That used to work before. Look at the current image where I have 1-10/12. but only 6 results due to changing the filter. 

image

jurajpiar:
> Initial pagination works fine, but the Pagination is not updated once you change the filter. That used to work before. Look at the current image where I have 1-10/12. but only 6 results due to changing the filter.
> 
> image

Done https://github.com/rsksmart/rif-marketplace-cache/pull/529#issuecomment-774681718

julianmrodri (approved):
LGTM

itofarina (approved):

PR #713, opened 2021-02-10
feat(tests): app context tests

jurajpiar:
Test coverage for the AppContext.

Creates `useProviderTest` utility HOC for simpler Context Provider testing

As part of the https://rsklabs.atlassian.net/browse/RMKT-457

### DIFF
--- a/src/__tests__/testUtils.tsx
+++ b/src/__tests__/testUtils.tsx
@@ -1,0 +1,28 @@
+import React, { FC, useContext } from 'react'
+
+const useProviderTest = <T extends object>(
+  Provider: React.ElementType,
+  context: React.Context<T>) => (
+    { test }: { test: Function },
+  ): React.ReactElement => {
+    const MockConsumer: FC = () => {
+      test(useContext(context))
+
+      return <div />
+    }
+
+    return (
+      <Provider>
+        <MockConsumer />
+      </Provider>
+    )
+  }
+
+test.skip('fake test', () => {
+  expect(true).toBeTruthy()
+})
+
+export default {}
+export {
+  useProviderTest,
+}
--- a/src/context/App/__tests__/AppContext.test.tsx
+++ b/src/context/App/__tests__/AppContext.test.tsx
@@ -1,6 +1,13 @@
 import { render } from '@testing-library/react'
+import { ServiceAddress } from 'api/models/serviceAddresses'
+import { confirmationAddress } from 'api/rif-marketplace-cache/blockchain/confirmations'
+import { serviceAddress as notificationsAddress } from 'api/rif-marketplace-cache/notifications'
+import { xrServiceAddress } from 'api/rif-marketplace-cache/rates/xr'
 import React, { FC, useContext } from 'react'
+import { storageAddresses } from 'api/rif-marketplace-cache/storage/interfaces'
+import { serviceAddress } from 'api/rif-storage-upload-service/upload/interfaces'
 import AppContext, { AppContextProvider } from '../AppContext'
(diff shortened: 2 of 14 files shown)


src/__tests__/testUtils.tsx
```diff
@@ -0,0 +1,24 @@
+import React, { FC, useContext } from 'react'
+
+const useProviderTest = <T extends object>(
```
farinalvaro, line 3: I would suggest to move this to another place, may be under `hooks/utilities`?
As it's used for tests but not running a specific test
This would also solve the failing test

src/context/App/__tests__/AppContext.test.tsx
```diff
@@ -1,55 +1,105 @@
...
+      })
+      describe('loaders', () => {
+        test(`should contain {
+              contract: false,
+              data: false,
+              filters: false,
+              other: false,
+            }`, () => {
+          render(<ProviderTest test={({ state: { loaders } }): void => {
+            expect(loaders).toBeTruthy()
```
farinalvaro, line 73: I think we should do 
```
const defaultLoaders = {
              contract: false,
              data: false,
              filters: false,
              other: false,
            }
toEqual(defaultLoaders)
```

if I understood correctly the idea of this second test.

src/context/App/__tests__/AppContext.test.tsx
```diff
@@ -1,55 +1,105 @@
...
+      describe('alertPanel', () => {
+        test(`should contain {
+            display: false,
+            message: '',
+          }`, () => {
+          render(<ProviderTest test={({ state: { alertPanel } }): void => {
+            expect(alertPanel).toEqual({
+              display: false,
+              message: '',
+            })
```
farinalvaro, line 98: may be extract object to const? like `defaultAlertPanel`

src/context/App/__tests__/appActions.test.tsx
```diff
@@ -0,0 +1,196 @@
...
+      })
+    })
+
+    test('should set state.messages.loading to { text: payload.message, type: \'info\' }', () => {
+      const payload: LoadingPayload = {
+        id: 'data',
+        isLoading: true,
+        message: 'Test message.',
+      }
+      expect(initialState.messages.loading).toBeFalsy()
```
farinalvaro, line 78: May be better to move the expect to the beginning of the test just to avoid confusion. (feel free to ignore 😅 )

src/context/App/__tests__/appActions.test.tsx
```diff
@@ -0,0 +1,196 @@
...
+      initialState.alertPanel.display = true
+      initialState.alertPanel.message = 'BS'
+
+      const {
+        alertPanel,
+      }: Partial<AppState> = HIDE_ALERT(initialState, {})
+      expect(alertPanel).toEqual({
+        display: false,
+        message: '',
+      })
```
farinalvaro, line 193: If extracted to const `defaultAlertPanel`, could be reused here

itofarina (commented):
Just some small comments but overall looks great!

src/__tests__/testUtils.tsx
jurajpiar: My reasoning for adding it under __tests__ is that nothing under this dir will be compiled for the browser. As this util is only needed in tests I'd rather keep it there. For the failing test as result of not having a single one in this file, I suggest I just add an empty test.

farinalvaro: Makes sense!

itofarina (approved):
Nice job!
I would only suggest to squash the last two commits if we are going to rebase

PR #716, opened 2021-02-11
feat(triggers-sell): adds route + tabs section with sell option

farinalvaro:
This PR is only to track the progress on the notifier feature and shouldn't be merged until we define our use of GitFlow or GitHubFlow


artem-brazhnikov (approved):
let's push it

Issue #397, opened 2020-10-09
rename errorReporterFactory to useErrorReporter or createErrorReporter


jurajpiar:
May be create a custom hook for this

PR #718, opened 2021-02-12
feat: adds useErrorReporter custom hook

farinalvaro:
Closes https://github.com/rsksmart/rif-marketplace-ui/issues/397

Adds ability to report errors easier with:

```
const errorReporter = useErrorReporter()

  errorReporter(new UIError({
    error: new Error(),
    id: 'some-id',
    text: 'some text',
  }))


jurajpiar (changes requested):
Nice one @farinalvaro!
Could you please also remove the errorReporterFactory from the AppContext module altogether and instead call the dispatch directly in the useCallback?

jurajpiar (approved):

PR #725, opened 2021-02-17
refactor(context): app and storage contexts

### LINKED ISSUE #330: Refactor Reducers
Refactor RNS and storage offers reducers to reducer per storage bases.

use following file structure:
\\/ contextFolder
- Context.tsx
- actions.ts
- interfaces.ts
- utils.ts
- index.ts

### index.ts
Where all definitions can keep their short, non-domain-specific names even when exported. The items necessary to be accessible from the outside of this namespace should be exported with using their domain-specific designations in the index.ts.

### interfaces.ts
use comments to show logical separation, such as:
```
// STATE
...
// PAYLOAD
...
// ACTIONS
...
```
#### Actions definition
First define Action types. This will force the dispatch call to follow the payload type, also allowing payload itself to be the desired value:
```
export type Action = (
  | {
    type: 'ACTION_TYPE_NAME'
    payload: { property: Type } | Type 
  } ...
)
```
Next, define type for the actions object to follow:
```
export type Actions = {
  ACTION_TYPE_NAME: (
    state: State,
    payload: { property: Type } | Type ,
  ) => State
```

Asynchronous actions should be carried within the context props in an object, sibling to state and dispatch, such as:
```
export type Props = {
    state: State
    dispatch: Dispatch<Action>
    asyncActions: AsyncActions
}
```

Finally, use `src/context/storage/buy/checkout/actions.ts` reducer function as template to try and change the reducer factory so that it doesn't need to be so heavily overtyped.

### LINKED ISSUE #396: Fix generic helper types
- src/context/storeUtils/interfaces.ts
These are meant to be used as type safety but as it is it doesn't provide much value due to inability to be extended.
Previously tried to resolve this with generics but it quickly became a mess, so it still awaits a solution.
This links to the typing of the reducer factory (src/context/storeUtils/reducer.ts), which uses some of these types and as a result is currently not type safe.

jurajpiar:
Quite a large PR but most files are not really changing anything and the things changed are all kinda linked so splitting it into separate PRs would be counterproductive. 
- removes unnecessary abstractions: ContextPayload, ContextReducer, ContextAction, ContextActions, MarketFilter
- removes deprecated storeReducerFactory
- removes superfluous TokenAddresses
- refactors contexts: Market/storage/offerEdit and App as these were the last two to use the deprecated storeReducerFactory and the other mentioned refactors
- the rest of the files just resolves the effects of the changes above 

Closes #330 
Closes #396


src/context/App/__tests__/appActions.test.tsx
```diff
@@ -1,14 +1,14 @@
...
+} from 'context/App/interfaces'
 import { LoaderId } from 'models/UIMessage'
-import { AppState, initialState } from '../AppContext'
+import { initialState, AppState, appActions } from '..'
 
 const testActionReturn = (
-  action: (state: AppState, payload: any) => AppState,
-  payload: AppPayload,
+  action: (state: AppState, payload: unknown) => AppState,
+  payload: unknown,
```
farinalvaro, line 11: Is it possible to specify better types?

src/models/marketItems/StorageFilters.ts
```diff
@@ -1,10 +1,9 @@
-import { MarketFilter } from 'models/Market'
 import { MinMaxFilter } from 'models/Filters'
 import { SubscriptionPeriod } from './StorageItem'
 
-export interface StorageOffersFilters extends MarketFilter {
-    price?: MinMaxFilter
-    size?: MinMaxFilter // total capacity. NOT available capacity!
+export interface StorageOffersFilters {
+    price: MinMaxFilter
+    size: MinMaxFilter // total capacity. NOT available capacity!
```
farinalvaro, line 6: Is there a particular reason for not allowing undefined values on these filter props?

itofarina (approved):
Amazing!

src/context/Services/rns/__tests__/OffersContext.test.tsx
```diff
@@ -51,7 +53,7 @@ describe('RnsOffersContextProvider', () => {
       renderTest(TestComponent)
     })
 
-    test('should contain boolean "needsRefresh" set to false', () => {
+    test.skip('should contain boolean "needsRefresh" set to false', () => {
```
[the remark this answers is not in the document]
jurajpiar: yea on purpose. it's failing and the test PR will fix it. Fixing it here would require more effort than is reasonable considering that.

src/context/App/__tests__/appActions.test.tsx
jurajpiar: well, yes and no. I removed the AppPayload as it is unnecessary in the app. I could combine the payloads here in the test, however it occurred to me that it isn't really necessary as it is just for tests. `unknown` is a valid type and if (and when) in the future we'd want to extract this fn to be used in other tests it will need to be `unknown` anyway.

src/models/marketItems/StorageFilters.ts
jurajpiar: yes, but actually no.) I'll change it back.

0xartem (approved):
Hi, the refactoring looks good to me. I have a couple questions about structure and style overall. But I think it's better for me to refresh the React knowledge before we talk about it. Didn't see any critical issues in terms of coding.

jurajpiar:
> Hi, the refactoring looks good to me. I have a couple questions about structure and style overall. But I think it's better for me to refresh the React knowledge before we talk about it. Didn't see any critical issues in terms of coding.

Let's talk about this in a call @artem-brazhnikov, ok?

PR #727, opened 2021-02-20
feat(tests-unit): adds storage upload and staking contexts

jurajpiar:
Part of the https://rsklabs.atlassian.net/browse/RMKT-457 ticket.


artem-brazhnikov:
can you check why CI is not passing?

jurajpiar:
> can you check why CI is not passing?

I think we have an issue with some memory leaks or some-such as the CI keeps running out of memory. All passed now.

0xartem (approved):

itofarina (approved):
LGTM

PR #732, opened 2021-03-01
fix(deps): removes tasegir and nfts

jurajpiar:
_Still a draft due to the issues described bellow, yet I thought might be worth to start reviewing this as there are many changes._

**Changes**:
- replaces rif-marketplace-nfts dependency with its static ABI
- removes tasegir and in its place adds commitlint (for ci) and [eslint-config-standard-with-typescript](https://www.npmjs.com/package/eslint-config-standard-with-typescript) (standard lib for linting in TS) and fixes resulting issues by both adding/disabling new rules to eslintrc and following the linter's suggestions
- updates react-scripts and webpack; this had also had an impact on the ease of removing tasegir

**Current issues**:
- the images loaded from rif-ui do not show
- ~some `confirmations` and `xr` api tests are failing~

**Otherwise left TODOs**:
- add commitlint
- set `@typescript-eslint/default-param-last` rule to **error** and fix the issues across the project
- set `@typescript-eslint/no-throw-literal` rule to **error** and fix the issue in `UIError`
- set `no-shadow` to **error** and fix the issues across the project


src/context/Services/notifications/Context.tsx
```diff
@@ -79,7 +79,7 @@ export const Provider: FC = ({ children }) => {
         }
       }
 
-      initialise()
+      initialise().finally(() => {})
```
artem-brazhnikov, line 82: can you explain what this line does?

0xartem (commented):
Please let me know how much time you think it will take you to fix currrent issues and TODOs? In hours.

jurajpiar:
The issue with the `jest.fn` not working as expected (i.e. not executing the callback given as argument) is rooted in one of the [breaking changes](https://github.com/facebook/create-react-app/blob/master/CHANGELOG.md#jest) of react-scripts v4. This sets the `resetMocks`to true as this seems to be the [best practice](https://github.com/facebook/create-react-app/pull/7899). However, as described in [Jest docs](https://github.com/facebook/create-react-app/pull/7899), resetting mocks after each test "does not restore their initial implementation".
The solution is to create the mock in ad-hoc fashion.

src/context/Services/notifications/Context.tsx
jurajpiar: this is a result of the no-floating-promises rule.

jurajpiar: this is a result of the [no-floating-promises](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-floating-promises.md) rule.

jurajpiar:
Closing as using PR https://github.com/rsksmart/rif-marketplace-ui/pull/742 instead, due to the issue with loading images resulting from the CRAv4.

PR #745, opened 2021-03-03
feat: add staging workflow for CD

rafaiovlabs:
Staging workflow for CD of marketplace UI

### DIFF
--- a/.github/workflows/staging.yml
+++ b/.github/workflows/staging.yml
@@ -1,0 +1,63 @@
+name: RIF Marketplace UI deploy
+
+on:
+  push:
+    tags:
+      - 'v[0-9]+.[0-9]+.[0-9]+-**'
+      - 'v[0-9]+.[0-9]+.[0-9]+'
+    branches:
+      - staging
+
+jobs:
+  deploy-staging:
+    name: RIF Marketplace UI - Staging
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2
+
+    - name: Setup Node 14
+      uses: actions/setup-node@v1
+      with:
+        node-version: '14'
+
+    - uses: actions/cache@v2
+      id: cache
+      with:
+        path: ~/.npm
+        key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
+        restore-keys: |
+          ${{ runner.os }}-node-
+
+    - name: Install dependencies
+      if: steps.cache.outputs.cache-hit != 'true'
+      run: |
+        npm ci --only=production
+
+    - name: Build site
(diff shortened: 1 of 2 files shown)


.github/workflows/staging.yml
```diff
@@ -0,0 +1,63 @@
+name: RIF Marketplace UI deploy
+
+on:
+  push:
+    tags:
+      - 'v[0-9]+.[0-9]+.[0-9]+-**'
+      - 'v[0-9]+.[0-9]+.[0-9]+'
+    branches:
+      - staging
```
artem-brazhnikov, line 9: ok good, the dev env is approved, I will work on the branches workflow so it's consistent across repos

artem-brazhnikov (approved):

jurajpiar (changes requested):
Change the commit message to follow the [conventional commit specs](https://www.conventionalcommits.org/en/v1.0.0/) please

artem-brazhnikov:
@jurajpiar @rafaiovlabs we need to find out why the pipeline is failing, I restarted the CI. There seemed to be an out of memory error.

artem-brazhnikov:
@rafaiovlabs can you please check the amount of memory allocated for the CI? I think it might not enough
https://discuss.circleci.com/t/unit-tests-enomem-not-enough-memory-read/35695/3

.github/workflows/staging.yml
```diff
@@ -0,0 +1,63 @@
+name: RIF Marketplace UI deploy
+
+on:
+  push:
+    tags:
```
artem-brazhnikov, line 5: @rafaiovlabs I was thinking through this and I am not sure we need to deploy tags to the staging. Image when we merged some piece of logic to staging and QA are working on it. And then we make a release from master and suddenly the QA are presented with a different version.

artem-brazhnikov (changes requested):
Also, in the previous PR there were jobs for the testnet and mainnet. Why they disappeared in this one?

.github/workflows/staging.yml
rafaiovlabs: @artem-brazhnikov  Fixed the mem issue of CI with this PR, about the tags..i thing those give us better control (in dev environment they are not needed at all) , but for staging could be of benefit and also we can have tags like v2.3.0-qafix or something like that. About not having CD for other envs thats correct i only have staging precisely to validate the deployment, workflows and so on, after validation CD for other envs is trivial to add.

artem-brazhnikov: ok, good

artem-brazhnikov: ok, lets try and see how it goes

artem-brazhnikov (approved):

PR #749, opened 2021-03-04
chore(release): v1.2.2

jurajpiar:


julianmrodri (commented):
Do we need to update Footer?

jurajpiar:
> Do we need to update Footer?

no that's updated via env variable that's set to whatever is in the package.json

itofarina (approved):

PR #753, opened 2021-03-05
feat(triggers): adds offers api

jurajpiar:
- offers service added to the the notifier folder
- offers service added to the apis list in the AppContext
- offers service defined the transport model for offers data
- NotifierOffersContext created connection with the offers service
- NotifierOffersContext fetched data from the service
- NotifierOffersContext stored listings

Resolves https://rsklabs.atlassian.net/browse/RMKT-585

### DIFF
--- a/src/api/rif-marketplace-cache/notifier/interfaces.ts
+++ b/src/api/rif-marketplace-cache/notifier/interfaces.ts
@@ -1,5 +1,4 @@
 import { APIService } from 'api/models/apiService'
-import { Modify } from 'utils/typeUtils'
 import { NotifierOfferItem } from 'models/marketItems/NotifierItem'
 import { NotifierOffersAddress, NotifierOffersWSChannel } from './offers'
 
@@ -8,11 +7,8 @@ export type Address =
 export type WSChannel =
     | NotifierOffersWSChannel
 
-export type NotifierAPIService = Modify<
-  APIService,
-  {
-    path: Address
-    _channel: WSChannel
-    fetch: () => Promise<NotifierOfferItem[]>
-  }
->
+export interface NotifierAPIService extends APIService {
+  path: Address
+  _channel: WSChannel
+  fetch: () => Promise<NotifierOfferItem[]>
+}
--- a/src/api/rif-marketplace-cache/notifier/interfaces.ts
+++ b/src/api/rif-marketplace-cache/notifier/interfaces.ts
@@ -1,0 +1,18 @@
+import { APIService } from 'api/models/apiService'
+import { Modify } from 'utils/typeUtils'
+import { NotifierOfferItem } from 'models/marketItems/NotifierItem'
+import { NotifierOffersAddress, NotifierOffersWSChannel } from './offers'
+
+export type Address =
+    | NotifierOffersAddress
+export type WSChannel =
+    | NotifierOffersWSChannel
+
+export type NotifierAPIService = Modify<
+  APIService,
+  {
+    path: Address
+    _channel: WSChannel
+    fetch: () => Promise<NotifierOfferItem[]>
+  }
+>
--- a/src/api/rif-marketplace-cache/notifier/offers/models.ts
+++ b/src/api/rif-marketplace-cache/notifier/offers/models.ts
@@ -1,15 +1,15 @@
 import { NotifierOfferItem, NotifierPlan } from 'models/marketItems/NotifierItem'
 
-export type TransportModelType = {
+export interface NotifierOffersTransport {
     provider: string
     plans: NotifierPlan[]
 }
-class TransportModel implements TransportModelType {
+class TransportModel implements NotifierOffersTransport {
     provider!: string
 
     plans!: NotifierPlan[]
 
-    constructor({ provider, plans }: TransportModelType) {
+    constructor({ provider, plans }: NotifierOffersTransport) {
       this.plans = plans
       this.provider = provider
     }
--- a/src/api/rif-marketplace-cache/notifier/offers/models.ts
+++ b/src/api/rif-marketplace-cache/notifier/offers/models.ts
@@ -1,0 +1,5 @@
+export class TransportModel {
+
+}
+
+export default {}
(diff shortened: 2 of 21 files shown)


src/api/rif-marketplace-cache/notifier/offers/models.ts
```diff
@@ -0,0 +1,23 @@
+import { NotifierOfferItem, NotifierPlan } from 'models/marketItems/NotifierItem'
+
+export type TransportModelType = {
+    provider: string
+    plans: NotifierPlan[]
+}
+class TransportModel implements TransportModelType {
```
artem-brazhnikov, line 7: if you are implementing it's some interface - it's better to use the `interface` keyword and not use the `Type` word in its name

src/api/rif-marketplace-cache/notifier/offers/models.ts
```diff
@@ -0,0 +1,23 @@
...
+    provider!: string
+
+    plans!: NotifierPlan[]
+
+    constructor({ provider, plans }: TransportModelType) {
+      this.plans = plans
+      this.provider = provider
+    }
+
+    public readonly toLocal = (): NotifierOfferItem => ({
```
artem-brazhnikov, line 17: you can just use a member function here I guess

src/api/rif-marketplace-cache/notifier/interfaces.ts
```diff
@@ -0,0 +1,18 @@
...
+import { Modify } from 'utils/typeUtils'
+import { NotifierOfferItem } from 'models/marketItems/NotifierItem'
+import { NotifierOffersAddress, NotifierOffersWSChannel } from './offers'
+
+export type Address =
+    | NotifierOffersAddress
+export type WSChannel =
+    | NotifierOffersWSChannel
+
+export type NotifierAPIService = Modify<
```
artem-brazhnikov, line 11: The same here, I would call it an `interface` if it has a contract that needs implemented

src/api/rif-marketplace-cache/notifier/offers/offers.ts
```diff
@@ -0,0 +1,37 @@
...
+export type Address = typeof address
+
+export const wsChannel = 'offers' as const
+export type WSChannel = typeof wsChannel
+
+class OffersService extends AbstractAPIService
+  implements NotifierAPIService {
+    path = address
+
+    private mapFromTransport: MapFromTransport<TransportModel, NotifierOfferItem> = (transport) => transport.toLocal()
```
artem-brazhnikov, line 18: can be expressed as a member function

src/components/pages/notifier/buy/NotifierOffersPage.tsx
```diff
@@ -40,41 +41,51 @@ const NotifierOffersPage: FC = () => {
...
       } = item
+
+      // FIXME: Refactor
+      const channels = Array.from(new Set(plans.flatMap((plan) => plan.channels))).join(', ')
+      const allCurrencis = plans.flatMap((plan) => plan.priceOptions.map((option) => SUPPORTED_TOKEN_RECORDS[option.token].displayName))
+      const allAmounts = plans.flatMap((plan) => plan.limit).sort()
+      const allPrices = plans.flatMap((plan) => plan.priceOptions.map((option) => {
+        const xrRate = crypto[option.token].rate
+        return option.value.mul(xrRate)
+      })).sort()
```
artem-brazhnikov, line 60: why don't you write a single cycle that will go over the plans and make all the necessary transformations in one go?

0xartem (changes requested):

src/api/rif-marketplace-cache/notifier/offers/models.ts
jurajpiar: fair enough. though since I'm not planning on extending this interface it should remain a type ([src](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#differences-between-type-aliases-and-interfaces))?
About the naming, yea I know. Wasn't sure what to call it tho. Got any suggestions?

src/api/rif-marketplace-cache/notifier/offers/models.ts
jurajpiar: Could you describe why it would be preferable?
I'm guessing you just mean to define it as:
```ts
class TransportModel {
   ...
   readonly toLocal() {
      ...
   }
}
```


In fact, this is something I've just tried out for this PR. Normally, we just use a type here and the mapping is left for the mapFromTransport function. The reason for this approach was that I thought it would be more easily testable once we refactor the base api class.
Anyways, I can and perhaps should return it back to match the rest of the app's implementation.

src/api/rif-marketplace-cache/notifier/interfaces.ts
jurajpiar: as you can observe from the link in the previous comment, the difference between `type` and `interface` in TS is really in the extensibility not implementability, thus it only make sense to use `interface` when you expect the type to be extended.

src/api/rif-marketplace-cache/notifier/offers/offers.ts
jurajpiar: again, please give reason. Just because it can doesn't mean it should.

src/components/pages/notifier/buy/NotifierOffersPage.tsx
jurajpiar: That's exactly why I asked the reviewers to focus on this bit. If you look more closely each bit has different requirements for the mapping. For example, the `channels` needs to put it all in a `Set` to remove duplications, the `prices` needs to return a mapped inner parameter, then just the first and the last is used, and so on.
This complexity is the exact reason I kept it all in separate loops, so that we can see it better and discuss the best strategy.

Actually I just noticed the `currency` is wrong. It should display all currency options, so would be more similar to the `channels` in that it needs to be a `Set`.

0xartem (approved):
LGTM

PR #773, opened 2021-03-16
fix(storage): adds check for crypto existence

jurajpiar:
... on my purchases page.

This should fix the bug reported in the ticket [RMKT-597](https://rsklabs.atlassian.net/browse/RMKT-597).

Not verified locally as this bug could not be locally reproduced.

### DIFF
--- a/src/components/organisms/storage/mypurchases/PurchasesTable.tsx
+++ b/src/components/organisms/storage/mypurchases/PurchasesTable.tsx
@@ -91,7 +91,7 @@ const PurchasesTable: FC<PurchasesProps> = (
     view: '',
   }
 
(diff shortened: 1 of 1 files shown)


artem-brazhnikov (approved):
approved

farinalvaro (approved):
Basically LGTM 👌
I would suggest to also report an error so users know what's going on when they don't see their purchases because of this

jurajpiar:
> Basically LGTM 👌
> I would suggest to also report an error so users know what's going on when they don't see their purchases because of this

Good idea, though this should be done on lower level so that it doesn't need to be thrown on each page separately. I'd suggest the context that manages the exchange rates. I'll create an issue for it.

PR #788, opened 2021-03-25
feat(devops): adding cd fix for staging branch

rafaiovlabs:
The CD workflow must be added to the proper branch in order to execute correctly.

### DIFF
--- a/.github/workflows/staging.yml
+++ b/.github/workflows/staging.yml
@@ -2,9 +2,6 @@ name: RIF Marketplace UI deploy
 
 on:
   push:
-    tags:
-      - 'v[0-9]+.[0-9]+.[0-9]+-**'
-      - 'v[0-9]+.[0-9]+.[0-9]+'
     branches:
       - staging
 
--- a/.github/workflows/staging.yml
+++ b/.github/workflows/staging.yml
@@ -33,7 +33,7 @@ jobs:
     - name: Install dependencies
       if: steps.cache.outputs.cache-hit != 'true'
       run: |
-        npm ci --only=production
+        npm ci
 
     - name: Build site
       env:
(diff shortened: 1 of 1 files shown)


artem-brazhnikov:
@rafaiovlabs I want to revisit one question we have already discussed here: https://github.com/rsksmart/rif-marketplace-ui/pull/745/files#r592971714. So imagine that the QA are testing on staging a new feature from the staging branch. But the master contains the code for release, like now for example. And as soon as we create a new tag - it will re-deploy the same version to the staging which will downgrade it and break QA experience. How do you think we can handle it? Maybe we should delete tags deployment for staging and add tags with postfix to testnet only?

rafaiovlabs (commented):
When this PR is merded the automatic deployment will work under these circumstances:
* if on "staging" branch a new tag is added with the form v #.#.# or v #.#.#-word. then the deploy will be automatic

rafaiovlabs:
> @rafaiovlabs I want to revisit one question we have already discussed here: https://github.com/rsksmart/rif-marketplace-ui/pull/745/files#r592971714. So imagine that the QA are testing on staging a new feature from the staging branch. But the master contains the code for release, like now for example. And as soon as we create a new tag - it will re-deploy the same version to the staging which will downgrade it and break QA experience. How do you think we can handle it? Maybe we should delete tags deployment for staging and add tags with postfix to testnet only?

I'm ok with with whatever workflow fits your needs. the use case you mention can be overcome by the development process itself but ok already changed it, we will have tags only for testnet then. regards.

artem-brazhnikov:
> > @rafaiovlabs I want to revisit one question we have already discussed here: https://github.com/rsksmart/rif-marketplace-ui/pull/745/files#r592971714. So imagine that the QA are testing on staging a new feature from the staging branch. But the master contains the code for release, like now for example. And as soon as we create a new tag - it will re-deploy the same version to the staging which will downgrade it and break QA experience. How do you think we can handle it? Maybe we should delete tags deployment for staging and add tags with postfix to testnet only?
> 
> I'm ok with with whatever workflow fits your needs. the use case you mention can be overcome by the development process itself but ok already changed it, we will have tags only for testnet then. regards.

Thank you, Rafael.

artem-brazhnikov (approved):

jurajpiar (approved):

PR #792, opened 2021-03-28
fix(devops): erasing not needed ref to script

rafaiovlabs:


jurajpiar:
I'm not sure I understand why it is not needed. After all, the `build:prod` script is the one that is going to build using the `tsconfig.prod.json` to pre compile without tests. When you said we don't use it when merging this change to staging I imagined you meant that that's because it does't contain the `tsconfig.prod.json` file. Otherwise I am not sure what is going to stop the deployment from failing upon seeing the devDependencies used in the code of the test files.
Am I missing something?

rafaiovlabs:
> @rafaiovlabs I'm not sure I understand why it is not needed. After all, the `build:prod` script is the one that is going to build using the `tsconfig.prod.json` to pre compile without tests. When you said we don't use it when merging this change to staging I imagined you meant that that's because it does't contain the `tsconfig.prod.json` file. Otherwise I am not sure what is going to stop the deployment from failing upon seeing the devDependencies used in the code of the test files.
> Am I missing something?

Thats correct, I'm just trying to fix the building can you add then the tsconfig.prod.json file to the staging branch and close this PR ?

jurajpiar:
> Thats correct, I'm just trying to fix the building can you add then the tsconfig.prod.json file to the staging branch and close this PR ?

Yea, sure can do. Thanks

PR #809, opened 2021-03-31
fix(rifui): removes muitypography styles override

jurajpiar:
- by copying affected components and all other necessary files
    for their correct run from the rif-ui
 - This should be reverted once these issues of typography style override
    is resolved in the rif-ui dependency!

This is just a workaround!
Bug described in RUI issue https://github.com/rsksmart/rif-ui/issues/237.


0xartem (approved):
approved

PR #811, opened 2021-04-01
feat(triggers): shows providers' plans

jurajpiar:
Resolves [RMKT-627](https://rsklabs.atlassian.net/browse/RMKT-627)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,4 @@ npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
 .vscode/
+tsconfig.json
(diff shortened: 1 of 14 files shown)


.gitignore
```diff
@@ -23,3 +23,4 @@ npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
 .vscode/
+tsconfig.json
```
0xartem, line 26: why is this one in gitignore?

0xartem (approved):
LGTM

.gitignore
jurajpiar: because this file is copied before run each time from either `tsconfig.prod.json` or `tsconfig.dev.json` so it shouldn't really exist in the repo.

PR #826, opened 2021-04-13
feat(triggers): adds My Offers page

jurajpiar:
Resolving https://rsklabs.atlassian.net/browse/RMKT-624


src/api/rif-marketplace-cache/notifier/subscriptions/index.ts
```diff
@@ -0,0 +1,16 @@
...
+export {
+  address as notifierSubscriptionsAddress,
+  wsChannel as notifierSubscriptionsWSChannel,
+}
+
+export default SubscriptionsService
+
+export type NotifierSubscriptionsAddress = Address
+export type NotifierSubscriptionsWSChannel = WSChannel
+export type NotifierSubscriptionsTransportModel = SubscriptionDTO
```
0xartem, line 16: what is the purpose of this file? it seems only to provide aliases. Also, can you explain to me why are you using those aliases for addresses and channels and the model. Are there other addresses that have different structures?

src/api/rif-marketplace-cache/notifier/interfaces.ts
```diff
@@ -1,11 +1,14 @@
...
 import { NotifierOfferItem } from 'models/marketItems/NotifierItem'
 import { NotifierOffersAddress, NotifierOffersWSChannel } from './offers'
+import { NotifierSubscriptionsAddress, NotifierSubscriptionsWSChannel } from './subscriptions'
 
 export type Address =
     | NotifierOffersAddress
+    | NotifierSubscriptionsAddress
 export type WSChannel =
     | NotifierOffersWSChannel
+    | NotifierSubscriptionsWSChannel
```
0xartem, line 11: it seems to export `Address` here but in the original file it is doing the opposite

src/api/rif-notifier-service/index.ts
```diff
@@ -0,0 +1,49 @@
...
+          text: `Error: Could not get subscriptions for account ${account}`,
+          id: 'service-fetch',
+        })
+      }
+
+      return response.json()
+    }
+
+   getSubscriptionInfo = async (subscriptionId: string):
+   Promise<Subscriptions[]> => {
```
0xartem, line 34: both methods return the Subscriptions array. maybe this one only needs a single one?

src/components/molecules/plans/PlanViewActions.tsx
```diff
@@ -0,0 +1,44 @@
+import Grid, { GridProps } from '@material-ui/core/Grid'
+import { Button } from '@rsksmart/rif-ui'
+import { ButtonProps } from '@material-ui/core/Button' // FIXME: this is meant to come from RUI, but @rsksmart/rif-ui/dist/components/atoms/Button module cannot be found. FIX RUI!
+// import Button, { ButtonProps } from '@rsksmart/rif-ui/dist/components/atoms/Button'
```
0xartem, line 4: can we remove these comments

src/components/pages/notifier/myoffers/mapActiveContracts.tsx
```diff
@@ -0,0 +1,100 @@
...
+  notifBalance: 'Notifications',
+  expDate: 'Expiration date',
+  price: 'Price',
+  actions: '',
+}
+export type ActiveContractItem = {
+    [K in keyof typeof activeContractHeaders]: React.ReactElement
+  } & { id: string}
+
+const mapActiveContracts = (
```
0xartem, line 26: can you decompose this one? not critical

0xartem (approved):
Left a couple of comments, looks good overall. Maybe a biggy one, need to think next time to divide such a story. If possible.

jurajpiar:
> Left a couple of comments, looks good overall. Maybe a biggy one, need to think next time to divide such a story. If possible.

Thanks for the review. You may remember that when we pointed the story we expected all the components to be ready from a previous (storage) story. However, when I started working on it I realised that those components were not reusable, ie they were coupled with the storage service.
Having said that, this was still a failure on my part of not double checking that those components were indeed reusable, rather than just assuming they would be.

src/api/rif-marketplace-cache/notifier/subscriptions/index.ts
jurajpiar: This is because internally, as in within this folder, the names do not have to be so specific, as everything here relates to one feature. However, in order to be used outside without confusion, the parts need to have domain-specific name.
Like all indexes, this one provides an entry-point to a feature, in this case it is api/notifier/subscriptions module. This pattern is used across the project.

src/api/rif-marketplace-cache/notifier/interfaces.ts
jurajpiar: yes, so, this is a level above, ie it is the interfaces definitions for the module api/notifier. Here the Address refers to an aggregation of notifier address types. And again, within the module it does not need to specify what this refers to. In the index file this type is then exported as NotifierServiceAddress to specify what it is for the outside of the module.

src/components/molecules/plans/PlanViewActions.tsx
jurajpiar: No, I wouldn't. Left them here intentionally to make sure this gets changed once rif-ui is fixed.

src/components/pages/notifier/myoffers/mapActiveContracts.tsx
jurajpiar: Do you mean the type definition or the customer object in the filter and the map? Anyway, I'll change both.

src/api/rif-marketplace-cache/notifier/subscriptions/index.ts
0xartem: okay, can you share with me some link to read more about this practice?

PR #832, opened 2021-04-17
feat(multicurrency): adds isWhitelistedToken check

jurajpiar:
Adds validation on choice of currencies. These have to:
1. exist in the config under the service's tokens key
2. be whitelisted by the contract


src/contracts/rns/Marketplace.ts
```diff
@@ -86,6 +89,34 @@ class MarketplaceContract extends ContractBase {
...
+
+    const wlByType = await this._call<{0: boolean, 1: boolean, 2: boolean}>(
+      this.methods.whitelistedPaymentToken(tokenAddress),
+      txOptions,
+    )
+
+    const wLTokens = {
+      [TOKEN_TYPES.ERC20]: wlByType[0],
+      [TOKEN_TYPES.ERC677]: wlByType[1],
+      [TOKEN_TYPES.ERC777]: wlByType[2],
```
0xartem, line 115: do we really support those tokens?

0xartem (approved):
code looks good, just 1 question

src/contracts/rns/Marketplace.ts
jurajpiar: ERC677 is RIF token
ERC777 is a standard like ERC20.
But the point here is that the contract's whitelistedPaymentToken method returns type { 0: boolean, 1: boolean, 2: boolean } where 0,1 and 2 are the represented tokens. So even tho we don't I thought we should capture them correctly.

PR #837, opened 2021-04-30
feat(notifier): adds checkout page

vjgene:
Add NotifierOffersSelectedPage
List events in the page
Provide way to delete events

### DIFF
--- a/src/components/organisms/notifier/NotifierPlanDescription.tsx
+++ b/src/components/organisms/notifier/NotifierPlanDescription.tsx
@@ -1,0 +1,75 @@
+/* eslint-disable @typescript-eslint/no-empty-function */
+import {
+  Grid, Typography,
+} from '@material-ui/core'
+import DescriptionCard from 'components/molecules/DescriptionCard'
+import RifAddress from 'components/molecules/RifAddress'
+import React, {
+  FC,
+} from 'react'
+import { OffersOrder } from 'context/Services/notifier/offers/interfaces'
+import { MarketCryptoRecord } from 'models/Market'
+
+type Props = {
+      item?: OffersOrder
+    currentFiat: string
+    crypto: MarketCryptoRecord
+}
+
+const NotifierPlanDescription: FC<Props> = ({ item, crypto, currentFiat }) => {
+  if (!item) return <DescriptionCard />
+  const { token, daysLeft, value } = item.item
+  return (
+    <DescriptionCard>
+      <Grid container spacing={1}>
+        <Grid item xs={1} md={2}>
+          <Typography color="textSecondary" noWrap>Provider</Typography>
+        </Grid>
+        <Grid item xs={1} md={2}>
+          <Typography color="textSecondary" noWrap>Notifications</Typography>
+        </Grid>
+        <Grid item xs={1} md={2}>
+          <Typography color="textSecondary" noWrap>Channels</Typography>
+        </Grid>
+        <Grid item xs={1} md={2}>
+          <Typography color="textSecondary" noWrap>Currencies</Typography>
+        </Grid>
+        <Grid item xs={1} md={2}>
+          <Typography color="textSecondary" noWrap>Price</Typography>
+        </Grid>
+        <Grid item xs={1} md={2}>
+          <Typography color="textSecondary" noWrap>Expiration Period</Typography>
+        </Grid>
+        <Grid item xs={1} md={2}>
+          <RifAddress value={item.item['provider']} color="textPrimary" noWrap />
+        </Grid>
+        <Grid item xs={1} md={2}>
+          <Typography color="textPrimary">{item.item['limit']}</Typography>
+        </Grid>
+        <Grid item xs={1} md={2}>
+          <Typography color="textPrimary">{item.item['channels'].join('+')}</Typography>
+        </Grid>
+        <Grid item xs={1} md={2}>
+          <Typography color="textPrimary">{item.item['token'].displayName}</Typography>
+        </Grid>
+        <Grid item xs={1} md={2}>
+          <Typography color="textPrimary">
+            {((crypto?.[token.symbol]?.rate || 0) * value.toNumber()).toPrecision(4)}
+            {' '}
+            {currentFiat}
+          </Typography>
+        </Grid>
+        <Grid item xs={1} md={2}>
+          <Typography color="textPrimary">
+            {daysLeft}
+            {' '}
+            day
+            {daysLeft !== 1 ? 's' : ''}
+          </Typography>
+        </Grid>
+      </Grid>
+    </DescriptionCard>
+  )
+}
+
+export default NotifierPlanDescription
--- a/src/components/pages/notifier/NotifierRoutes.tsx
+++ b/src/components/pages/notifier/NotifierRoutes.tsx
@@ -11,6 +11,7 @@ import { NotifierOffersContextProvider } from 'context/Services/notifier/offers'
 import NotifierLandingPage from './NotifierLandingPage'
 import { NotFound } from '..'
 import NotifierSellPage from './sell/NotifierSellPage'
+import NotifierOffersSelectedPage from './buy/NotifierOffersSelectedPage'
 import NotifierOffersPage from './buy/NotifierOffersPage'
 import NotifierMyOffersPage from './myoffers/NotifierMyOffersPage'
 import { buildTabs } from '../routerUtils'
@@ -66,6 +67,11 @@ const NotifierRoutes: FC = () => {
                     exact
                     path={ROUTES.NOTIFIER.BUY.BASE}
                     component={NotifierOffersPage}
+                  />
+                  <Route
+                    exact
+                    path={ROUTES.NOTIFIER.BUY.CHECKOUT}
+                    component={NotifierOffersSelectedPage}
                   />
                   {DeadEndRoute}
                 </Switch>
--- a/src/components/pages/notifier/buy/NotifierOffersSelectedPage.tsx
+++ b/src/components/pages/notifier/buy/NotifierOffersSelectedPage.tsx
@@ -1,0 +1,116 @@
+import React, {
+  FC, useContext, useState,
+} from 'react'
+
+import {
+  Grid,
+  Typography,
+} from '@material-ui/core'
+import { NotifierOffersContextProps as ContextProps, NotifierOffersContext } from 'context/Services/notifier/offers'
+import { Button } from '@rsksmart/rif-ui'
+import Marketplace from 'components/templates/marketplace/Marketplace'
+import CenteredPageTemplate from 'components/templates/CenteredPageTemplate'
+import NotifierPlanDescription from 'components/organisms/notifier/NotifierPlanDescription'
+import MarketContext, { MarketContextProps } from 'context/Market'
+import TableContainer from '@material-ui/core/TableContainer'
+import { Item } from 'models/Market'
+import RemoveButton from 'components/atoms/RemoveButton'
+import Tooltip from '@material-ui/core/Tooltip'
+
+const eventHeaders = {
+  name: 'Name',
+  type: 'Type',
+  channels: 'Channels',
+  actions: '',
+} as const
+
+type EventItem = Item & {
+  [K in keyof Omit<typeof eventHeaders, 'actions'>]: string
+} & {
+    signature: string
+}
+
+const NotifierOffersSelectedPage: FC = () => {
+  const {
+    state: {
+      exchangeRates: {
+        currentFiat: {
+          displayName: currentFiat,
+        },
+        crypto,
+      },
+    },
+  } = useContext<MarketContextProps>(MarketContext)
+
+  const {
+    state: {
+      order,
+    },
+    dispatch,
+  } = useContext<ContextProps>(NotifierOffersContext)
+
+  /* TODO use real events instead of hard coded events */
+  const [events, setEvents] = useState<Array<EventItem>>(
+    [
+      {
+        id: 'event1',
+        name: 'event1',
+        type: 'custom',
+        signature: 'test1',
+        channels: 'API',
+      },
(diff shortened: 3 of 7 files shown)


0xartem (approved):
looks good to me, wait for @jurajpiar review before merging

src/components/pages/notifier/buy/NotifierOffersSelectedPage.tsx
```diff
@@ -0,0 +1,129 @@
...
+          </MUITableRow>
+
+        </TableBody>
+        {/* Header */ }
+        <Grid item xs={11} md="auto">
+          <Typography gutterBottom variant="h6" color="primary">
+            Notification events added
+          </Typography>
+        </Grid>
+      </Table>
```
jurajpiar: Please, use the `src/components/molecules/DescriptionCard.tsx`. Also, use `Grid` instead of `Table`, as that renders quite a different result and it's a bit overkill for a single raw. (See `NotifierMyOffersPage` and `NotifierProviderDescription`)

jurajpiar: Also, it is missing styling (via `Typography`). Again check the the mentioned components for reference.

src/components/pages/notifier/buy/NotifierOffersSelectedPage.tsx
```diff
@@ -0,0 +1,129 @@
...
+                    iconButtonProps={{ onClick: () => { deleteEvent(index) } }}
+                    tooltipTitle="Delete event"
+                  />
+                </TableCell>
+              </MUITableRow>
+
+            </>
+          ))}
+        </TableBody>
+      </Table>
```
jurajpiar: When inserting a table, please use the Marktetplace component. For reference, see `src/components/organisms/plans/PlanView.tsx`. Also missing Typography as well.

src/context/Services/notifier/offers/Context.tsx
```diff
@@ -9,8 +9,10 @@ import { UIError } from 'models/UIMessage'
 import React, {
   createContext, FC, useContext, useEffect, useReducer, useState,
 } from 'react'
+import Big from 'big.js'
 import actions from './actions'
 import { Props, State } from './interfaces'
+import { SUPPORTED_TOKEN_RECORDS } from '../../../../contracts/interfaces'
```
jurajpiar: prefer absolute paths

src/components/pages/notifier/buy/NotifierOffersSelectedPage.tsx
```diff
@@ -0,0 +1,129 @@
...
+import {
+  Grid,
+  Table, TableBody, TableCell, TableHead, TableRow as MUITableRow,
+  Typography,
+} from '@material-ui/core'
+import { NotifierOffersContextProps as ContextProps, NotifierOffersContext } from 'context/Services/notifier/offers'
+import { Button, TooltipIconButton } from '@rsksmart/rif-ui'
+import ClearIcon from '@material-ui/icons/Clear'
+import { TableHeaders } from '../../../templates/marketplace/Marketplace'
+import CenteredPageTemplate from '../../../templates/CenteredPageTemplate'
```
jurajpiar: prefer absolute paths

src/context/Services/notifier/offers/__tests__/Context.test.tsx
```diff
@@ -1,8 +1,11 @@
 import React from 'react'
 import { render } from '@testing-library/react'
 import { useProviderTest } from '__tests__/testUtils'
+import Big from 'big.js'
 import { NotifierOffersContextProvider, NotifierOffersContext } from '..'
-import { OffersListing } from '../interfaces'
+import { OffersListing, OffersOrder } from '../interfaces'
+
+import { SUPPORTED_TOKEN_RECORDS } from '../../../../../contracts/interfaces'
```
jurajpiar: prefer absolute paths

src/context/Services/notifier/offers/Context.tsx
```diff
@@ -43,6 +45,38 @@ export const initialState: State = {
...
+        channels: 'API',
+      },
+      {
+        name: 'event2',
+        type: 'custom',
+        signature: 'test2',
+        channels: 'API',
+      },
+    ],
+  },
```
jurajpiar: no need for these here, as they'll only be used on this page. Move to local state, please.

src/context/Services/notifier/offers/Context.tsx
```diff
@@ -43,6 +45,38 @@ export const initialState: State = {
...
+      channels: [],
+      daysLeft: 0,
+      provider: '',
+      limit: 0,
+      token: SUPPORTED_TOKEN_RECORDS.rbtc,
+      value: new Big(1),
+      url: '',
+    },
+    isProcessing: true,
+  },
```
jurajpiar: why are you initialising the order? By default there can't be one, before you select something.

src/context/Services/notifier/offers/actions.ts
```diff
@@ -31,6 +31,18 @@ const actions: Actions = {
...
+    const newEvents = events.items.filter(
+      (x, y) => y !== row.index,
+    )
+    return {
+      ...state,
+      events: {
+        items: newEvents,
+      },
+    }
+  },
```
jurajpiar: there is no need for this as (un)setting the order should be sufficient. One order will, by design, always have only one item. See the the `ServiceOrder` type in `src/context/Services/interfaces.ts`.

src/models/marketItems/NotifierItem.ts
```diff
@@ -22,6 +22,13 @@ export type NotifierPlan = Provider & {
     daysLeft: number
 }
 
+export type NotifierEvent = {
+    name: string
+    type: string
+    signature: string
+    channels: string
```
jurajpiar: shouldn't this be `string[]`?

src/routes.ts
```diff
@@ -1,5 +1,6 @@
 const LISTING = 'listing'
 const CHECKOUT = 'checkout'
+const SELECTED = 'selected'
```
jurajpiar: remove and use checkout instead

jurajpiar (changes requested):
Please, see the `NotifierMyOffersPage` for reference to the components required for this page.

src/models/marketItems/NotifierItem.ts
vjgene: since the channels attribute are used for display only, it is of string type

src/components/organisms/notifier/NotifierPlanDescription.tsx
```diff
@@ -0,0 +1,79 @@
...
+  daysLeft: 'Expiration Period',
+}
+
+type Props = {
+      item: NotifierOrder
+    currentFiat: string
+    crypto: MarketCryptoRecord
+}
+
+export type NotifierOrder = OffersOrder | undefined | boolean
```
jurajpiar: I don't understand this type. Doesn't look very safe, being both object and a boolean that is.
Why can't the item prop be defined just as `item?: OffersOrder`?

src/components/organisms/notifier/NotifierPlanDescription.tsx
```diff
@@ -0,0 +1,79 @@
...
+      </Grid>
+    ))
+}
+      {
+        Object.keys(headers).map((itemName: string) => (
+          <Grid item xs={1} md={2}>
+            {renderProviderItem(itemName, item, crypto, currentFiat)}
+          </Grid>
+        ))
+}
```
jurajpiar: Seems totally unnecessary, considering these fields never change.
You could just put them down statically. No need for two loops and a function with lots of if-statements.

src/components/pages/notifier/NotifierRoutes.tsx
```diff
@@ -81,6 +87,8 @@ const NotifierRoutes: FC = () => {
                     component={NotifierMyOffersPage}
                   />
                   {DeadEndRoute}
+
+                  <Route component={NotFound} />
```
jurajpiar: Remove this line as it is covered by the DeadEndRoute constant on the previous line.

src/context/Services/notifier/offers/interfaces.ts
```diff
@@ -11,6 +11,9 @@ export type ContextName = typeof contextName
 export type OffersListing = {
   items: NotifierOfferItem[]
 }
+export type NotifierEvents = {
+  items: NotifierEvent[]
+}
```
jurajpiar: Remove as not used in this context.

src/context/Services/notifier/offers/interfaces.ts
```diff
@@ -21,6 +24,7 @@ export type State = Modify<ServiceState<NotifierOfferItem>, {
   order?: OffersOrder
   filters: NotifierOffersFilters
   limits: ContextLimits
+  events?: NotifierEvents
```
jurajpiar: Remove from Context. It should only be in a local (in-page) state.

src/models/marketItems/NotifierItem.ts
```diff
@@ -22,6 +23,14 @@ export type NotifierPlan = Provider & {
...
+    actions: ReactElement
```
jurajpiar: Remove actions from this type. You can extend it locally on the page if you need to add it, for example
`const [events, setEvents] = useState<NotifierEvents & { actions: React.ReactElement }>()`.

But that's still kinda wrong, because the items you pass to the table should all be components, rather than text, so that they can be individually styled. So, I'd suggest to remove this type from here altogether and instead type it on the page like so:
```
const eventHeaders = {
  name: 'Name',
  type: 'Type',
  signature: 'Signature',
  channels: 'Channels',
  actions: '',
}

type EventTableItem = Item & {
    [K in keyof typeof eventHeaders]: React.ReactElement
}
const [events, setEvents] = useState<Array<EventTableItem>>()

```

src/components/pages/notifier/buy/NotifierOffersSelectedPage.tsx
```diff
@@ -0,0 +1,115 @@
...
+      ],
+    },
+  )
+
+  const removeEvent = (e) => {
+    events.items = events.items.filter((i) => i.name !== e.currentTarget.id)
+    setEvents({ ...events })
+  }
+
+  const RemoveButton: FC<string> = (id) => <Typography align="right" gutterBottom color="primary"><SelectRowButton size="large" variant="text" handleSelect={removeEvent} id={id}>X</SelectRowButton></Typography>
```
jurajpiar: 1. please, refactor this out into atoms folder
2. the 'X' should be replaced with `CloseIcon` imported from `@material-ui/icons/Close`
2. the `id` needs to be enclosed in curly brackets, but
3. the type should change to `FC<SelectRowButtonProps>`, that way you can spread the props to the `SelectRowButton`

src/components/pages/notifier/buy/NotifierOffersSelectedPage.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+  const collection: MarketplaceItem[] = events.items
+    .map<MarketplaceItem>((event) => ({
+      id: event.name,
+      provider: 0,
+      ...event,
+      actions: RemoveButton(event.name),
+    }))
```
jurajpiar: Seems unnecessary. You've already got the `events` constant that will satisfy the collection prop.
On the other hand, if you don not want to store components in the state (probably best not to), then you can do:
```
const eventHeaders = {
  name: 'Name',
  type: 'Type',
  signature: 'Signature',
  channels: 'Channels',
  actions: '',
} as const

type EventHeaders = typeof eventHeaders

type EventItem = Item & {
    [K in keyof Omit<EventHeaders, 'actions'>]: string
}

const [events, setEvents] = useState<Array<EventItem>>()

type EventTableItem = Item & {
   [K in keyof EventHeaders]: React.ReactElement
}

const collection = events.items
    .map<EventTableItem>((event) => ({
   ...
```

src/components/pages/notifier/buy/NotifierOffersSelectedPage.tsx
jurajpiar: then in the mapping you'd use pass prop 
```
handleSelect={() => {
   const newEvents = [ ...events ]
   newEvents.splice(i, 1)
   setEvents(newEvents)
}}
```

src/components/pages/notifier/buy/NotifierOffersSelectedPage.tsx
```diff
@@ -0,0 +1,115 @@
...
+    events.items = events.items.filter((i) => i.name !== e.currentTarget.id)
+    setEvents({ ...events })
+  }
+
+  const RemoveButton: FC<string> = (id) => <Typography align="right" gutterBottom color="primary"><SelectRowButton size="large" variant="text" handleSelect={removeEvent} id={id}>X</SelectRowButton></Typography>
+
+  const collection: MarketplaceItem[] = events.items
+    .map<MarketplaceItem>((event) => ({
+      id: event.name,
+      provider: 0,
```
jurajpiar: there is no provider key

src/components/pages/notifier/buy/NotifierOffersSelectedPage.tsx
```diff
@@ -0,0 +1,115 @@
...
+          actions: <br />,
+        },
+        {
+          name: 'event2',
+          type: 'custom',
+          signature: 'test2',
+          channels: 'API',
+          actions: <br />,
+        },
+      ],
```
jurajpiar: As I mentioned further down, either store values and map them to components, or store components so you don't have to map. Personally, I think the first one is better as storing components seems evil .]

jurajpiar (changes requested):
A few more changes, sorry.

jurajpiar (approved):
looking good, make sure to clean the commits before merging .]

Issue #710, opened 2021-02-08
Standardize data inputs to use react-hook-form


farinalvaro:
https://react-hook-form.com/ provides an easy and robust way to input data. We are starting to use it in #709. 

It would be great to propagate the use of this library for data inputs

PR #842, opened 2021-05-10
feat(notifier): adds notification channels

vjgene:
Adds Notification channels section in notifier buy checkout page

### DIFF
--- a/src/components/organisms/notifier/NotificationChannel.tsx
+++ b/src/components/organisms/notifier/NotificationChannel.tsx
@@ -1,0 +1,71 @@
+import React, { FC } from 'react'
+import { makeStyles, Theme } from '@material-ui/core/styles'
+import Grid from '@material-ui/core/Grid'
+import Typography from '@material-ui/core/Typography'
+import Box from '@material-ui/core/Box'
+import { colors, shortenString } from '@rsksmart/rif-ui'
+import { NotifierChannel } from 'models/marketItems/NotifierItem'
+import RemoveButton from 'components/atoms/RemoveButton'
+
+export interface NotificationChannelProps {
+  className?: string
+  onRemoveClick: (e: any) => void
+  channel: NotifierChannel
+}
+
+const useStyles = makeStyles((theme: Theme) => ({
+  innerContainer: {
+    backgroundColor: colors.transparent,
+    borderRadius: 10,
+    padding: theme.spacing(2),
+    border: `1px solid ${colors.gray3}`,
+    height: '25%',
+  },
+  leftContent: {
+    [theme.breakpoints.up('sm')]: {
+      borderRight: `1px solid ${colors.gray3}`,
+    },
+  },
+}))
+
+const NotificationChannel: FC<NotificationChannelProps> = ({
+  className = '', channel, onRemoveClick,
+}) => {
+  const classes = useStyles()
+
+  return (
+    <Grid className={className} container alignItems="center" spacing={2}>
+      <Grid item xs={9}>
+        <Grid
+          container
+          alignItems="center"
+          className={classes.innerContainer}
+          spacing={1}
+        >
+          <Grid xs={3} item>
+            <Typography component="div">
+              <Box fontWeight="fontWeightMedium" textAlign="center" color={colors.gray5}>
+                {channel.type}
+              </Box>
+            </Typography>
+          </Grid>
+          <Grid xs={6} item style={{ display: 'flex', justifyContent: 'space-around', alignItems: 'center' }}>
+            {shortenString(channel.destination, 30, 25)}
+          </Grid>
+          <Grid xs={3} item>
+            <RemoveButton
+              style={{
+                maxWidth: '20px', maxHeight: '20px', minWidth: '20px', minHeight: '20px',
+              }}
+              id={channel.type}
+              handleSelect={onRemoveClick}
+            />
+          </Grid>
+        </Grid>
+      </Grid>
+      <Grid item xs={3} />
+    </Grid>
+  )
+}
+
+export default NotificationChannel
--- a/src/components/organisms/notifier/NotificationChannel.tsx
+++ b/src/components/organisms/notifier/NotificationChannel.tsx
@@ -1,11 +1,10 @@
 import React, { FC } from 'react'
-import { makeStyles, Theme } from '@material-ui/core/styles'
 import Grid from '@material-ui/core/Grid'
 import Typography from '@material-ui/core/Typography'
 import Box from '@material-ui/core/Box'
 import { colors, shortenString } from '@rsksmart/rif-ui'
 import { NotifierChannel } from 'models/marketItems/NotifierItem'
-import RemoveButton from 'components/atoms/RemoveButton'
+import RemovableRow from 'components/organisms/RemovableRow'
 
 export interface NotificationChannelProps {
   className?: string
@@ -13,59 +12,21 @@ export interface NotificationChannelProps {
   channel: NotifierChannel
 }
 
-const useStyles = makeStyles((theme: Theme) => ({
-  innerContainer: {
-    backgroundColor: colors.transparent,
-    borderRadius: 10,
-    padding: theme.spacing(2),
-    border: `1px solid ${colors.gray3}`,
-    height: '25%',
-  },
-  leftContent: {
-    [theme.breakpoints.up('sm')]: {
-      borderRight: `1px solid ${colors.gray3}`,
-    },
-  },
-}))
-
 const NotificationChannel: FC<NotificationChannelProps> = ({
-  className = '', channel, onRemoveClick,
-}) => {
-  const classes = useStyles()
-
-  return (
-    <Grid className={className} container alignItems="center" spacing={2}>
-      <Grid item xs={9}>
-        <Grid
-          container
-          alignItems="center"
-          className={classes.innerContainer}
-          spacing={1}
-        >
-          <Grid xs={3} item>
-            <Typography component="div">
-              <Box fontWeight="fontWeightMedium" textAlign="center" color={colors.gray5}>
-                {channel.type}
-              </Box>
-            </Typography>
-          </Grid>
-          <Grid xs={6} item style={{ display: 'flex', justifyContent: 'space-around', alignItems: 'center' }}>
-            {shortenString(channel.destination, 30, 25)}
-          </Grid>
-          <Grid xs={3} item>
-            <RemoveButton
-              style={{
-                maxWidth: '20px', maxHeight: '20px', minWidth: '20px', minHeight: '20px',
-              }}
-              id={channel.type}
-              handleSelect={onRemoveClick}
-            />
-          </Grid>
-        </Grid>
-      </Grid>
-      <Grid item xs={3} />
+  channel, onRemoveClick,
+}) => (
+  <RemovableRow id={channel.type} handleSelect={onRemoveClick}>
+    <Grid xs={3} item>
+      <Typography component="div">
+        <Box fontWeight="fontWeightMedium" textAlign="center" color={colors.gray5}>
+          {channel.type}
+        </Box>
+      </Typography>
     </Grid>
-  )
-}
+    <Grid xs={6} item style={{ display: 'flex', justifyContent: 'space-around', alignItems: 'center' }}>
+      {shortenString(channel.destination, 30, 25)}
+    </Grid>
+  </RemovableRow>
+)
 
 export default NotificationChannel
--- a/src/components/organisms/notifier/NotificationChannelCreate.tsx
+++ b/src/components/organisms/notifier/NotificationChannelCreate.tsx
@@ -1,0 +1,70 @@
+import React, { FC } from 'react'
+import { makeStyles, Theme } from '@material-ui/core/styles'
+import Grid from '@material-ui/core/Grid'
+import { NotifierChannel, NotifierChannelType } from 'models/marketItems/NotifierItem'
+import { FormControl } from '@material-ui/core'
+import Select from '@material-ui/core/Select'
+import MenuItem from '@material-ui/core/MenuItem'
+import TextField from '@material-ui/core/TextField'
+import { SelectRowButton } from 'components/molecules'
+
+export interface NotificationChannelCreateProps {
+  onChannelChange: (e: any) => void
+  onDestinationChange: (e: any) => void
+  notifierChannel: NotifierChannel
+  channelAdd: () => void
+    availableChannels: string[]
+}
+
+const useStyles = makeStyles((theme: Theme) => ({
+  formControl: {
+    margin: theme.spacing(1),
+    minWidth: 120,
+  },
+  select: {
+    textAlign: 'center',
+    height: 40,
+  },
+}))
+
+const NotificationChannelCreate: FC<NotificationChannelCreateProps> = ({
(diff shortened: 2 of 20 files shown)


src/components/organisms/notifier/NotificationChannel.tsx
```diff
@@ -0,0 +1,71 @@
...
+          </Grid>
+          <Grid xs={3} item>
+            <RemoveButton
+              style={{
+                maxWidth: '20px', maxHeight: '20px', minWidth: '20px', minHeight: '20px',
+              }}
+              id={channel.type}
+              handleSelect={onRemoveClick}
+            />
+          </Grid>
```
itofarina: Please, prefer classes instead of inline styles. See [this SO question](https://stackoverflow.com/questions/57791954/material-ui-why-should-i-use-makestyles-instead-of-inline-styles) for more info

src/components/organisms/notifier/NotificationChannel.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+  return (
+    <Grid className={className} container alignItems="center" spacing={2}>
+      <Grid item xs={9}>
+        <Grid
+          container
+          alignItems="center"
+          className={classes.innerContainer}
+          spacing={1}
+        >
```
itofarina: If this is the same row styles we were talking about, I think it would be best to template it. That way we could reuse it here and if we want to change it in the future we would only need to do the change in one place.

src/components/organisms/notifier/NotificationChannelsList.tsx
```diff
@@ -0,0 +1,124 @@
...
+    const newChannels = addedChannels.filter((i) => i.type !== e.currentTarget.id)
+    setChannels(newChannels)
+  }
+
+  const collection = addedChannels.map((channel) => ({
+    type: channel.type,
+    destination: channel.destination,
+  }))
+
+  return (
```
itofarina: I know we have a lot of `Grid` components acting as forms, but we are trying to migrate them to https://react-hook-form.com/. With that we don't need to create validations for emails for example. It also improves UX.
See https://github.com/rsksmart/rif-marketplace-ui/issues/710 and https://github.com/rsksmart/rif-marketplace-ui/pull/709 for reference

itofarina (commented):
Overall looks good. Just please consider those 2 comment:
- Template the component we wanted to reuse instead of copying it
- Use `react-hook-form` library for data input

src/models/marketItems/NotifierItem.ts
```diff
@@ -29,3 +29,13 @@ export type NotifierSubscriptionItem = Item & Omit<SubscriptionDTO, 'hash' | 'pr
   price: Big
   token: SupportedToken
 }
+
+export enum NotifierChannelType {
+   API = 'Enter api destination',
+   EMAIL = 'Enter your email'
+}
```
jurajpiar: Prefer using const assertion to enums, unless you'd benefit from bidirectional resolution. See https://tinyurl.com/enumconst

src/utils/validationUtils.ts
```diff
@@ -0,0 +1,18 @@
+export const validateURL = (_url: string): boolean => {
+  let url
+  try {
+    url = new URL(_url)
+  } catch (_) {
+    return false
+  }
+  return url.protocol === 'http:' || url.protocol === 'https:'
```
jurajpiar: what about other protocols?

src/utils/validationUtils.ts
```diff
@@ -0,0 +1,18 @@
...
+}
+
+export const validateEmail = (emails: string): boolean => {
+  const re = /^(([^<>()[\]\\.,;:\s@"]+(\.[^<>()[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/
+  let valid = true
+  emails.split(';').forEach((email) => {
+    if (re.test(String(email).toLowerCase()) === false) valid = false
+  })
+  return valid
+}
```
jurajpiar: let's not validate emails in the UI.
1. expensive
2. doesn't actually validate the email address

once we do start using email in channels, we should send a verification email instead.

src/components/pages/notifier/buy/NotifierOffersSelectedPage.tsx
```diff
@@ -109,6 +110,9 @@ const NotifierOffersSelectedPage: FC = () => {
       <Button variant="outlined" color="primary" rounded>
         + Add Notification Events
       </Button>
+      <br />
+      <br />
```
jurajpiar, line 121: Use grid to position things on page instead.

src/components/organisms/notifier/NotificationChannelsList.tsx
```diff
@@ -0,0 +1,124 @@
...
+      />
+      <Grid className={classes.channelsList} item xs={12}>
+        <Grid alignItems="center" container spacing={2}>
+          {
+          collection.map((channel) => <NotificationChannel className={classes.notificationChannel} channel={channel} onRemoveClick={removeChannel} />)
+        }
+        </Grid>
+      </Grid>
+      <br />
+      <br />
```
jurajpiar: don't need this. use grid and/or css

src/components/organisms/notifier/NotificationChannelsList.tsx
```diff
@@ -0,0 +1,124 @@
...
+        notifierChannel={notifierChannel}
+        availableChannels={availableChannels}
+        onDestinationChange={(e) => setDestination(e.currentTarget.value)}
+        onChannelChange={handleChange}
+        channelAdd={addChannel}
+      />
+      <Grid className={classes.channelsList} item xs={12}>
+        <Grid alignItems="center" container spacing={2}>
+          {
+          collection.map((channel) => <NotificationChannel className={classes.notificationChannel} channel={channel} onRemoveClick={removeChannel} />)
```
jurajpiar: split this to lines

src/components/organisms/notifier/NotificationChannelsList.tsx
```diff
@@ -0,0 +1,124 @@
...
+    const index = addedChannels.findIndex((ch) => ch.type === notifierChannel.type)
+
+    if (index === -1) addedChannels.push(notifierChannel)
+    else addedChannels[index].destination = destination
+    const newChannels = [...addedChannels]
+    setChannels(newChannels)
+  }
+
+  const removeChannel = (e): void => {
+    const newChannels = addedChannels.filter((i) => i.type !== e.currentTarget.id)
```
jurajpiar: please use meaningful variable names (yea, even for iterator arguments)

src/components/organisms/notifier/NotificationChannelsList.tsx
```diff
@@ -0,0 +1,124 @@
...
+    return false
+  }
+
+  const addChannel = () => {
+    notifierChannel.destination = destination
+
+    if (!validate(notifierChannel)) {
+      return
+    }
+    const index = addedChannels.findIndex((ch) => ch.type === notifierChannel.type)
```
jurajpiar: please use meaningful variable names (yea, even for iterator arguments)

src/components/organisms/notifier/NotificationChannelsList.tsx
```diff
@@ -0,0 +1,124 @@
...
+      }
+    }
+    return false
+  }
+
+  const addChannel = () => {
+    notifierChannel.destination = destination
+
+    if (!validate(notifierChannel)) {
+      return
```
jurajpiar: use the forms dependency for validation, if needed. (would not worry about that right now tbh tho)

src/components/organisms/notifier/NotificationChannelsList.tsx
```diff
@@ -0,0 +1,124 @@
...
+      type: e.target.value as string,
+      destination: '',
+    })
+  }
+  const classes = useStyles()
+
+  if (!channels) {
+    return <Grid />
+  }
+  const availableChannels: Array<string> = channels.map((ch) => (NotifierChannelType[ch] ? ch : ''))
```
jurajpiar: please use meaningful variable names (yea, even for iterator arguments)

src/components/organisms/notifier/NotificationChannelsList.tsx
```diff
@@ -0,0 +1,124 @@
...
+const NotificationChannelsList: FC<Props> = ({ channels }) => {
+  const [notifierChannel, setChannel] = useState<NotifierChannel>({
+    type: channels ? channels[0] : '',
+    destination: '',
+  })
+  const [addedChannels, setChannels] = useState<Array<NotifierChannel>>([])
+  const [destination, setDestination] = useState<string>('')
+  const handleChange = (e) => {
+    setChannel({
+      type: e.target.value as string,
```
jurajpiar: destructure {target} or {target: {value}}

src/components/organisms/notifier/NotificationChannelsList.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+type Props =
+    {
+    channels: string[] | undefined
+}
+
+const NotificationChannelsList: FC<Props> = ({ channels }) => {
+  const [notifierChannel, setChannel] = useState<NotifierChannel>({
+    type: channels ? channels[0] : '',
+    destination: '',
```
jurajpiar: I'd keep default undefined and render the `NotificationChannelCreate` conditionally

jurajpiar (changes requested):
A couple more changes from me too, please.

src/utils/validationUtils.ts
vjgene: Notifier only supports http and https for rest calls as of now

src/utils/validationUtils.ts
```diff
@@ -0,0 +1,11 @@
+const validateURL = (_url: string): boolean => {
+  let url
+  try {
+    url = new URL(_url)
+  } catch (_) {
+    return false
+  }
+  return url.protocol === 'http:' || url.protocol === 'https:'
```
jurajpiar: I think it would be better to use a constant for the valid protocols. Also, we're trying to keep the use of mutable variables close to none. Another point is to try not to use arg names with underscore prefix. In fact I'm surprised eslint didn't complain about that. I suggest following structure:

```suggestion
const EVENT_API_AVAILABLE_PROTOCOLS = [ 'http:', 'https:'] // we could potentially make this configurable, for sure we should move it somewhere more in scope of the notifier service

const validateURLProtocol = (url: string, validProtocols: Array<string>): boolean => {
  try {
    const { protocol } = new URL(url)
    return validProtocols.includes(protocol);
  } catch { // note that catch in js doesn't need arguments
    return false
  }
}
```

src/components/organisms/notifier/NotificationChannelsList.tsx
```diff
@@ -0,0 +1,105 @@
...
+      type: value,
+      destination: '',
+    })
+  }
+  const classes = useStyles()
+
+  if (!channels) {
+    return <Grid />
+  }
+  const availableChannels: Array<string> = channels.map((channel) => (notifierChannelType[channel] ? channel : ''))
```
jurajpiar: I get that you want to restrict the use to only those currently supported channels with this. If so, then we should make this either configurable, or a constant somewhere. However, one that represents this, like `const EVENT_AVAILABLE_CHANNELS = ['API']` (currently just that API I think. Is that right @artem-brazhnikov?).
Another thing here is that maybe use `.filter` instead of mapping to potentially empty string.

src/components/organisms/notifier/NotificationChannelCreate.tsx
```diff
@@ -0,0 +1,92 @@
...
+import { FormControl } from '@material-ui/core'
+import Select from '@material-ui/core/Select'
+import MenuItem from '@material-ui/core/MenuItem'
+import TextField from '@material-ui/core/TextField'
+import { SelectRowButton } from 'components/molecules'
+import { useForm } from 'react-hook-form'
+import Typography from '@material-ui/core/Typography'
+import validateURL from 'utils/validationUtils'
+
+export interface NotificationChannelCreateProps {
```
jurajpiar: As not used outside just call it `Props`

src/models/marketItems/NotifierItem.ts
```diff
@@ -29,3 +29,13 @@ export type NotifierSubscriptionItem = Item & Omit<SubscriptionDTO, 'hash' | 'pr
   price: Big
   token: SupportedToken
 }
+
+export const notifierChannelType = {
```
jurajpiar: Sorry didn't pick up on this before, but let's call this `notifierChannelPlaceholder` and move it to the `src/components/organisms/notifier/NotificationChannelCreate.tsx ` as it is not needed for any other areas of the UI.

jurajpiar (changes requested):
A couple more bits & bobs.

src/components/organisms/notifier/NotificationChannelCreate.tsx
```diff
@@ -0,0 +1,105 @@
...
+        {
+              errors.destination && (
+              <Typography color="error" variant="caption">
+                Invalid destination
+              </Typography>
+              )
+          }
+      </Grid>
+      <Grid item xs={3} md={3}>
+        <SelectRowButton id="add-channel" handleSelect={handleSubmit(() => channelAdd(notifierChannel))} size="large" variant="outlined">Add Channel</SelectRowButton>
```
jurajpiar: split to multiple lines

src/components/organisms/notifier/NotificationChannelsList.tsx
```diff
@@ -0,0 +1,95 @@
...
+    marginBottom: theme.spacing(4),
+    marginLeft: theme.spacing(15),
+    borderTop: `1px solid ${colors.gray3}`,
+    padding: theme.spacing(3, 0),
+  },
+
+}))
+
+type Props = {
+    channels: string[] | undefined
```
jurajpiar: - `channels?: Array< SupportedEventChannel>`

src/components/organisms/notifier/NotificationChannelsList.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+const NotificationChannelsList: FC<Props> = ({ channels }) => {
+  const [addedChannels, setChannels] = useState<Array<NotifierChannel>>([])
+
+  const classes = useStyles()
+
+  if (!channels) {
+    return <Grid />
+  }
+  const availableChannels: Array<string> = channels.filter((channel) => SUPPORTED_EVENT_CHANNELS.indexOf(channel as SupportedEventChannel) !== -1)
```
jurajpiar: - multiple lines
- `const availableChannels: Array< SupportedEventChannel>`
- SUPPORTED_EVENT_CHANNELS.includes(channel)

src/utils/validationUtils.ts
```diff
@@ -0,0 +1,10 @@
+export const EVENT_API_AVAILABLE_PROTOCOLS = ['http:', 'https:']
```
jurajpiar: this should be in the notifier config now. (also could you please rename it to match the other const please? my bad, sorry)

jurajpiar (approved):

PR #844, opened 2021-05-13
fix(devops): add proper network id for dev

rafaiovlabs:


0xartem (approved):

.github/workflows/develop.yml
```diff
@@ -40,8 +40,8 @@ jobs:
         REACT_APP_UPLOAD_ADDR: ${{ secrets.DEVELOP_REACT_APP_UPLOAD_ADDR }}
         REACT_APP_LOG_LEVEL: error
         REACT_APP_NETWORK: rskdevnet
-        REACT_APP_REQUIRED_NETWORK_ID: 31
-        REACT_APP_REQUIRED_NETWORK_NAME: RSK Develop
+        REACT_APP_REQUIRED_NETWORK_ID: 33
+        REACT_APP_REQUIRED_NETWORK_NAME: RSK Devnet
```
lucasvuotto: `RSK Devnet` is defined as such in `rskj` code, but the network ID in that case is `32`. I think it would be best if this was named `RSK RegTest` or `Marketplace Devnet` or something else that can't create the confussion between the expected environments in the future.

rafaiovlabs: Ok changed to RSK Regtest

jurajpiar (approved):

PR #846, opened 2021-05-14
fix: removes conditional call to setState hook

itofarina:
The app is failing to compile because of `setState` hook being called conditionally. This PR fixes it.

Screen Shot 2021-05-14 at 10 55 15


vjgene (approved):
Removing this condition is also an option as this condition is not necessary.

0xartem (approved):

jurajpiar (approved):
Ah yea, that was my fault. sorry. thanks for the fix

PR #848, opened 2021-05-16
feat(notifier): adds funds withdraw as provider

jurajpiar:
Resolves https://rsklabs.atlassian.net/browse/RMKT-626

Depends on https://github.com/rsksmart/rif-marketplace-cache/pull/661

### DIFF
--- a/src/components/templates/ProgressOverlay.tsx
+++ b/src/components/templates/ProgressOverlay.tsx
@@ -45,7 +45,7 @@ const ProgressOverlay: FC<Props> = ({
                 {
                   buttons.map((btn) => (
                     <GridItem
-                      key={btn.key || (btn as unknown as HTMLElement).innerText}
+                      key={String(btn.props.children)}
                     >
                       {btn}
                     </GridItem>
@@ -62,3 +62,4 @@ const ProgressOverlay: FC<Props> = ({
 }
 
 export default ProgressOverlay
+export type ProgressOverlayProps = Props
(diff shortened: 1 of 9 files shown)


src/contracts/notifier/Notifier.ts
```diff
@@ -57,6 +60,28 @@ class NotifierContract extends ContractWithTokens {
...
   }
 
+  public async withdrawFunds({
+    hash, token: { tokenAddress }, amount, from,
+  }: {
+    hash: string
+    token: BaseToken
+    amount: Big
+    from: string
+  }, txOptions: TxOptions = {}): Promise<TransactionReceipt> {
```
0xartem: let's make it a `type`

0xartem (approved):
overall looks good

src/components/templates/ProgressOverlay.tsx
```diff
@@ -45,7 +45,7 @@ const ProgressOverlay: FC<Props> = ({
                 {
                   buttons.map((btn) => (
                     <GridItem
-                      key={btn.key || (btn as unknown as HTMLElement).innerText}
```
itofarina, line 48: Why would we stop using `btn.key` prop when available here?

itofarina (commented):
Apart from that small comment LGTM

src/components/templates/ProgressOverlay.tsx
jurajpiar: ah yea..tx

PR #866, opened 2021-05-26
feat(notifier): adds provider validation

vjgene:
Adds provider url validation during registration


src/models/UIMessage.ts
```diff
@@ -11,7 +12,8 @@ export interface CustomAction {
     action: Function
 }
 
-export type ErrorId = APIErrorId | ContractErrorId | MarketErrorId
+export type ErrorId = APIErrorId | ContractErrorId |
+    MarketErrorId | NotifierErrorId
```
jurajpiar: Remove from here. It's too domain specific for this place.

src/components/organisms/notifier/ProviderRegistrar.tsx
```diff
@@ -23,6 +26,21 @@ const ProviderRegistrar: FC<ProviderRegistrarProps> = ({
...
   const { register, handleSubmit, errors } = useForm<Inputs>()
+  const reportError = useErrorReporter()
+
+  const validateProviderURL = async (url: string): Promise<boolean> => {
+    const notifierService: ProviderValidationService = new ProviderValidationService(url)
+    notifierService.connect(reportError)
+    const result: string = await notifierService.healthCheck().catch((error) => {
+      reportError({
+        error,
+        id: 'notifier-provider',
```
jurajpiar: that's just an api error type. the text should be used to narrow down the location of the error.

src/api/rif-notifier-service/provider.ts
```diff
@@ -0,0 +1,27 @@
...
+
+      if (content.length) return message
+      throw new UIError({
+        error: new Error('No available subscription plan'),
+        text: 'Atleast one subscription plan must be available for registration',
+        id: 'service-fetch',
+      })
+    }
+    return API_RESPONSE_MESSAGES.ERROR
+  }
```
jurajpiar: how come you didn't just add and endpoint to the existing api file?

jurajpiar (dismissed):
Tiny things though I feel I you might have misunderstood me when I said to just use the existing service for your endpoint.

src/api/rif-notifier-service/provider.ts
vjgene: This i wanted to separate validation logic from core notification service

jurajpiar (approved):
LGTM

src/components/organisms/notifier/ProviderRegistrar.tsx
```diff
@@ -52,7 +70,10 @@ const ProviderRegistrar: FC<ProviderRegistrarProps> = ({
                   inputProps={{
                     style: { textAlign: 'center' },
                   }}
-                  inputRef={register({ required: true })}
+                  inputRef={register({
+                    required: true,
+                    validate: { validateProviderURL },
```
itofarina, line 67: 😍

src/constants/strings.ts
```diff
@@ -10,4 +10,9 @@ and still undergoing final testing before the official release.
   and any purchasing of storage space should be done
    at the user’s own risk.`
 
+export const API_RESPONSE_MESSAGES = {
+  OK: 'OK',
+  ERROR: 'error',
+}
```
itofarina, line 16: Could it be better to have it as
`type ApiResponse = 'OK' | 'ERROR'`?

src/components/organisms/notifier/ProviderRegistrar.tsx
```diff
@@ -23,6 +26,21 @@ const ProviderRegistrar: FC<ProviderRegistrarProps> = ({
   providerAddress, onRegister, isEnabled,
 }) => {
   const { register, handleSubmit, errors } = useForm<Inputs>()
+  const reportError = useErrorReporter()
+
+  const validateProviderURL = async (url: string): Promise<boolean> => {
+    const notifierService: ProviderValidationService = new ProviderValidationService(url)
+    notifierService.connect(reportError)
+    const result: string = await notifierService.healthCheck().catch((error) => {
```
itofarina: I would prefer to not use string return and use a type instead like:
`result: ApiResponse = ...`

itofarina (approved):
Apart from the minor `type` comments looks great!

src/constants/strings.ts
jurajpiar: either that or better still,
```suggestion
export const API_RESPONSE_STATUS = {
  OK: 'OK',
  ERROR: 'error',
}

export type ApiResponse = keyof typeof API_RESPONSE_STATUS
```

src/components/organisms/notifier/ProviderRegistrar.tsx
jurajpiar: if you are going to do apply this suggestion Vijay, then you might as well do it properly and move the logic part to a mapping.
Also I've just realised that the service is misused a bit. Let's have a call about it.

jurajpiar (changes requested):

jurajpiar (approved):

PR #872, opened 2021-06-02
feat(notifier): edit provider

vjgene:
Adds ability to edit a given provider after a warning message


src/components/organisms/notifier/offers/ProviderEdit.tsx
```diff
@@ -0,0 +1,206 @@
...
+                  As a result of this operation, if the new url contains a different set of subscription plans than the current set of plans in marketplace, the current marketplace plans will be set to inactive, and any subscriptions that belong to those plans will be affected.
+                </Typography>
+              </Box>
+              <Box pt={4}>
+                <RoundBtn onClick={confirmRegistration}>Confirm</RoundBtn>
+              </Box>
+
+            </Grid>
+          </RifDialog>
+        ) }
```
itofarina: I think it could be cleaner to have all the `RifDialog`  instance and its content defined outside of the return method or in a separated component, as it changes a lot what's being rendered.  What do you think?

src/components/organisms/notifier/offers/ProviderEdit.tsx
```diff
@@ -0,0 +1,206 @@
...
+import RifDialog from 'components/organisms/RifDialog'
+import Box from '@material-ui/core/Box'
+import Grid from '@material-ui/core/Grid'
+import Staking from 'components/organisms/notifier/Staking'
+
+export type Props = {
+  mode: 'EDIT' | 'CREATE'
+}
+
+const ProviderEdit: FC<Props> = ({ mode }) => {
```
itofarina:
```suggestion
const ProviderEdition: FC<Props> = ({ mode }) => {
```
or 
```suggestion
const ProviderCreation: FC<Props> = ({ mode }) => {
```

src/components/organisms/notifier/offers/ProviderEdit.tsx
```diff
@@ -0,0 +1,206 @@
...
+import { useHistory } from 'react-router-dom'
+import ROUTES from 'routes'
+import RifDialog from 'components/organisms/RifDialog'
+import Box from '@material-ui/core/Box'
+import Grid from '@material-ui/core/Grid'
+import Staking from 'components/organisms/notifier/Staking'
+
+export type Props = {
+  mode: 'EDIT' | 'CREATE'
+}
```
itofarina: If there are only 2 modes may be we could directly use a boolean prop like `editMode` so that we don't have to be comparing every time

src/components/organisms/notifier/offers/ProviderEdit.tsx
```diff
@@ -0,0 +1,206 @@
...
+            contractAction: 'NOTIFIER_REGISTER_PROVIDER',
+            txHash: registerReceipt.transactionHash,
+          },
+        })
+      }
+    } catch (error) {
+      reportError({
+        error,
+        id: 'contract-notifier',
+        text: 'Could not register as a provider',
```
itofarina: When in edit mode this text should be `Could not edit...` right?

src/components/organisms/notifier/offers/ProviderEdit.tsx
```diff
@@ -0,0 +1,206 @@
...
+
+  const confirmRegistration = async (): Promise<void> => {
+    await handleRegistration({ endpointUrl: providerUrl })
+  }
+
+  return (
+    <CenteredPageTemplate>
+      <InfoBar
+        isVisible={Boolean(hasPendingConfs)}
+        text="Awaiting confirmations for new offer"
```
itofarina:
```suggestion
        text="Awaiting confirmations for provider registration"
```

src/components/organisms/notifier/offers/ProviderEdit.tsx
```diff
@@ -0,0 +1,206 @@
...
+            </Grid>
+          </RifDialog>
+        ) }
+      <Typography gutterBottom variant="h5" color="primary">
+        {mode === 'EDIT' ? 'Edit your information as provider' : 'Register as notifications provider'}
+      </Typography>
+      <Typography gutterBottom color="secondary" variant="subtitle1">
+        {mode === 'EDIT' ? 'Fill out the fields below to edit your trigger offer.' : 'Fill out the fields below to list your notification service.'}
+        All the information provided is meant to be true and correct.
+      </Typography>
```
itofarina: I think it could be better to define those strings outside like 
```const title = editMode? 'Edit...' : 'Register...'```

and just render `{title}` directly here.

Same for subtitle

File: src/components/organisms/notifier/offers/ProviderEdit.tsx [hunk 7]
```diff
@@ -0,0 +1,206 @@
...
+            buttonLabel="Edit Offer"
+          />
+        )
+        : (
+          <ProviderRegistrar
+            providerAddress={accountStr}
+            isEnabled={isWhitelistedProvider}
+            onRegister={handleRegistration}
+          />
+        )}
```
itofarina: Is it possible to check for edit mode inside the functions and act based on that when needed instead of conditionally rendering here? I would suggest to try that because the render is almost the same.

For example:
```

const buttonLabel = editMode?? "Edit offer"
...
return ...

          <ProviderRegistrar
            providerAddress={accountStr}
            isEnabled={isWhitelistedProvider}
            onRegister={({ endpointUrl }): void => {
              setShowConfirmProviderEdit(true)
              setProviderUrl(endpointUrl)
            }}
            buttonLabel={buttonLabel}
          />
```

src/components/organisms/notifier/offers/ProviderEdit.tsx
```diff
@@ -0,0 +1,206 @@
...
+      <ProgressOverlay
+        title="Registering as a provider!"
+        doneMsg="You have been registered as a provider!"
```
itofarina: When on edit mode:

```suggestion
        title="Editing your provider info!"
        doneMsg="Your provider info has been updated!"
```

src/components/pages/notifier/myoffers/NotifierMyOffersPage.tsx
```diff
@@ -125,11 +128,14 @@ const NotifierMyOffersPage: FC = () => {
...
+  const editProvider = (): void => {
+    history.push(ROUTES.NOTIFIER.MYOFFERS.EDIT)
+  }
+
   const handleAddPlan = logNotImplemented('handle add plan')
   const handleEditPlan = logNotImplemented('handle edit plan')
   const handleCancelPlan = logNotImplemented('handle cancel plan')
-  const handleEditProfile = logNotImplemented('handle edit profile')
-
+  const handleEditProfile = editProvider
```
itofarina: I think we could do this directly here
```suggestion
  const handleEditProfile = ():void => {
    history.push(ROUTES.NOTIFIER.MYOFFERS.EDIT)
  }
```

src/components/pages/notifier/sell/NotifierSellPage.tsx
```diff
@@ -1,151 +1,8 @@
...
-            View offers listing
-          </RoundBtn>,
-        ]}
-
-      />
-
-    </CenteredPageTemplate>
-  )
-}
+const NotifierSellPage: FC = () => <ProviderEdit mode="CREATE" />
```
itofarina: If this is the only thing we would do here I would suggest to remove this file 😄

src/components/pages/notifier/myoffers/NotifierEditOffersPage.tsx
```diff
@@ -0,0 +1,6 @@
+import React, { FC } from 'react'
+import ProviderEdit from 'components/organisms/notifier/offers/ProviderEdit'
+
+const NotifierEditOffersPage: FC = () => <ProviderEdit mode="EDIT" />
```
itofarina: May be remove this file if this is the only thing we need

itofarina (changes requested):
The idea looks good, I just think some things could be done simpler 😄. 
I'm happy to have a chat about that

src/components/pages/notifier/NotifierRoutes.tsx
```diff
@@ -97,7 +104,14 @@ const NotifierRoutes: FC = () => {
             <Route
               exact
               path={ROUTES.NOTIFIER.SELL.BASE}
-              component={NotifierSellPage}
+              component={
+        WithLoginCard({
+          WrappedComponent: () => <ManageProviderPage create />,
+          title: 'Connect your wallet to register as a provider',
+          contentText: 'Please, connect your wallet in order to register as a provider.',
+        })
```
jurajpiar: Use the `WithLoginCard` on the page, not in the router. Just make 2 routes here: one going to `NotifierSellPage` the other `ProviderEditPage` if you need to route to it.

src/components/organisms/notifier/provider/EditProvider.tsx
```diff
@@ -0,0 +1,68 @@
...
+            <Typography gutterBottom color="error">Your existing subscriptions and plans will be affected</Typography>
+            <Typography gutterBottom color="secondary">
+              As a result of this operation, if the new url contains a different set of subscription plans than the current set of plans in marketplace, the current marketplace plans will be set to inactive, and any subscriptions that belong to those plans will be affected.
+            </Typography>
+          </Box>
+          <Box pt={4}>
+            <RoundBtn onClick={(): Promise<void> => onRegister(providerUrl)}>Confirm</RoundBtn>
+          </Box>
+        </Grid>
+      </RifDialog>
```
jurajpiar: extract component.

src/components/organisms/notifier/provider/EditProvider.tsx
```diff
@@ -0,0 +1,68 @@
...
+import RifDialog from 'components/organisms/RifDialog'
+import ProviderRegistrar from '../ProviderRegistrar'
+
+type Props = {
+  isWhitelistedProvider: boolean
+  accountStr: string
+  onRegister: (string) => Promise<void>
+}
+
+const EditProvider: FC<Props> = ({
```
jurajpiar: Actually I think it would be best to just compose a dialog for the entire thing from the `myOffers` page and show the warning as a persistent element within the dialog. That way you don't need to touch the router, or to do conditional page composition. just render the warning with the form bellow it in a modal.

src/components/pages/notifier/provider/ManageProviderPage.tsx
```diff
@@ -87,70 +66,107 @@ const NotifierSellPage: FC = () => {
...
+            <AddProvider isWhitelistedProvider={isWhitelistedProvider} accountStr={accountStr} onRegister={handleRegistration}>
+              <ProgressOverlay
+                title="Registering as a provider!"
+                doneMsg="You have been registered as a provider!"
+                inProgress={processingTx}
+                isDone={txOperationDone}
+                buttons={[
+                  <RoundBtn
+                    key="go_to_my_offers"
+                    onClick={
```
jurajpiar: this is not good. consider my above comment to use dialog instead. Alternatively just use 2 different paths in the router. don't conditionally compose the entirety of a page.

jurajpiar (changes requested):
Getting there. Please consider and let me know what you think about my suggestions.

src/components/organisms/notifier/provider/EditProvider.tsx
vjgene: > Actually I think it would be best to just compose a dialog for the entire thing from the `myOffers` page and show the warning as a persistent element within the dialog. That way you don't need to touch the router, or to do conditional page composition. just render the warning with the form bellow it in a modal.

vjgene: Already myoffers page is too long, so we don't want to add more code. Also it's more clear to separate the functionality into EditProvider and AddProvider components since that's their function.

src/components/pages/notifier/NotifierRoutes.tsx
vjgene: > Use the `WithLoginCard` on the page, not in the router. Just make 2 routes here: one going to `NotifierSellPage` the other `ProviderEditPage` if you need to route to it.

This will result in two small pages which don't do much, which @farinalvaro doesn't agree with.

src/components/pages/notifier/provider/ManageProviderPage.tsx
vjgene: > this is not good. consider my above comment to use dialog instead. Alternatively just use 2 different paths in the router. don't conditionally compose the entirety of a page.

To avoid conditional, we can pass the components as children. But this will result in two pages which almost don't do much. Even react docs support conditional rendering https://reactjs.org/docs/conditional-rendering.html
also long as there are not too many conditions i suppose.

src/components/organisms/notifier/provider/EditProvider.tsx
vjgene: > extract component.

I think it's good idea. may be we can call it WarningDialog which takes some params. If we see a reusability

src/components/organisms/notifier/provider/EditProvider.tsx
jurajpiar: the more code would be effectively a single line, like `<EditProviderUrlDialog open={isChangingUrl}>`.

src/components/organisms/notifier/provider/EditProvider.tsx
itofarina: Resolving as it's not going to be reused for now

src/components/organisms/notifier/provider/EditProvider.tsx
itofarina: Resolving due to change on structure

itofarina (approved):
LGTM - team work 😃

src/components/organisms/notifier/provider/ProviderEdition.tsx
```diff
@@ -0,0 +1,152 @@
...
+    }
+  }
+
+  const buttonLabel = editMode ? 'Edit Offer' : undefined
+  const txDoneMsg = editMode
+    ? 'You provider info has been updated!'
+    : 'You have been registered as a provider!'
+  const txInProgressTitle = editMode
+    ? 'Editing your provider info!'
+    : 'Registering as a provider!'
```
vjgene: How about if we do this?

```
type labelTypes = 'BUTTON_LABEL' | 'INPROGRESS_MESSAGE' | 'DONE_MESSAGE'
const editLabels: Record<labelTypes, string > = {
  BUTTON_LABEL: 'Edit Offer',
  INPROGRESS_MESSAGE: 'Editing your provider info!',
  DONE_MESSAGE: 'Your provider info has been updated!',
}
const createLabels: Record<labelTypes, string> = {
  BUTTON_LABEL: 'Register Provider',
  INPROGRESS_MESSAGE: 'Registering as a provider!',
  DONE_MESSAGE: 'You have been registered as a provider!',
}

```
and then

```
const labels = editMode ? editLabels : createLabels
const buttonLabel = labels.BUTTON_LABEL
const txDoneMsg = labels.DONE_MESSAGE
const txInProgressTitle = labels.INPROGRESS_MESSAGE
```

vjgene (commented):
Looking good. I just think we can use editMode === true once to get all the labels

jurajpiar (approved):

src/components/organisms/notifier/provider/ProviderEdition.tsx
itofarina: Done, thanks

PR #874, opened 2021-06-04
feat(notifier): avoids multiple offers creation per provider

itofarina:
## Features
- [x] Add wrapper to sell page and enable to register only if check with cache that there are no offers as provider from the current account
- [x] Show spinner while fetching offers for current provider
- [x] Show "no multiple offers" card for notifier provider with redirection to your offer for edition
## Demo

https://user-images.githubusercontent.com/8449567/121590602-e24d4600-ca0e-11eb-9322-4176a1c9bb6f.mov

### DIFF
--- a/src/api/rif-marketplace-cache/notifier/offers/__tests__/api.test.ts
+++ b/src/api/rif-marketplace-cache/notifier/offers/__tests__/api.test.ts
@@ -2,6 +2,8 @@ import { AbstractAPIService } from 'api/models/apiService'
 import mockFeathersService from 'api/test-utils/feathers'
 import { NotifierOfferItem } from 'models/marketItems/NotifierItem'
 import { SYSTEM_SUPPORTED_SYMBOL } from 'models/Token'
+import { NotifierOffersFilters } from 'models/marketItems/NotifierFilters'
+import { SYSTEM_SUPPORTED_FIAT } from 'models/Fiat'
 import OffersService, { notifierOffersAddress, NotifierOffersTransportModel, notifierOffersWSChannel } from '../index'
 import { mapFromTransport } from '../api'
 import { NotifierOffersService } from '../..'
@@ -33,6 +35,15 @@ const MOCK_ITEM_0: NotifierOffersTransportModel = {
 const MOCK_RESPONSE: NotifierOffersTransportModel[] = [MOCK_ITEM_0]
 
 let offersService: NotifierOffersService
+
+const filters: NotifierOffersFilters = {
+  price: {
+    min: 1,
+    max: 10,
+    fiatSymbol: SYSTEM_SUPPORTED_FIAT.usd,
+  },
+  size: { min: 1, max: 10 },
+}
 
 describe('Notifier Offers Service', () => {
   beforeEach(() => {
@@ -77,12 +88,12 @@ describe('Notifier Offers Service', () => {
     test('should call service find method', async () => {
       const serviceFindSpy = jest.spyOn(offersService.service, 'find')
 
-      await offersService.fetch()
+      await offersService.fetch(filters)
       expect(serviceFindSpy).toBeCalled()
     })
 
     test('should return NotifierOfferItem[] on success', async () => {
-      const actualOffers: NotifierOfferItem[] = await offersService.fetch()
+      const actualOffers: NotifierOfferItem[] = await offersService.fetch(filters)
 
       const expectedOffers: NotifierOfferItem[] = MOCK_RESPONSE
         .map(mapFromTransport)
--- a/src/api/rif-marketplace-cache/notifier/offers/api.ts
+++ b/src/api/rif-marketplace-cache/notifier/offers/api.ts
@@ -37,7 +37,7 @@ export const mapFromTransport = ({
   limit: quantity,
   priceOptions: prices.map((price) => ({
     token: getSupportedTokenByName(
-        price.rateId as SupportedTokenSymbol,
+      price.rateId as SupportedTokenSymbol,
     ),
     value: parseToBigDecimal(price.price),
   })),
@@ -57,13 +57,13 @@ class OffersService extends AbstractAPIService
   _fetch = async (
     filters: NotifierOffersFilters,
   ): Promise<NotifierOfferItem[]> => {
-    const result: Paginated<PlanDTO> = await this.service.find({
-      query: filters && {
-        ...filters,
-        currency: Array.from(filters.currency),
-        channels: Array.from(filters.channels),
-      },
-    })
+    const { currency, channels } = filters
+    const query = filters && {
+      ...filters,
+      currency: currency && Array.from(currency),
+      channels: channels && Array.from(channels),
+    }
+    const result: Paginated<PlanDTO> = await this.service.find({ query })
 
     const { data, ...metadata } = isResultPaginated(result)
       ? result : { data: result }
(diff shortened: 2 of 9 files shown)


src/components/pages/notifier/sell/PageWrapper.tsx
```diff
@@ -0,0 +1,62 @@
...
+        provider: account,
+      })
+      setHasOffer(Boolean(currentOwnOffers[0]))
+      setIsLoadingOffer(false)
+    }
+
+    getOwnOffers()
+  }, [apis, account, reportError])
+
+  if (hasOffer) {
```
vjgene: Can we make another condition that if provider tries to register the same url that it shows a message you are already registered as a provider ?

src/api/rif-marketplace-cache/notifier/offers/__tests__/api.test.ts
```diff
@@ -77,12 +88,12 @@ describe('Notifier Offers Service', () => {
...
+      await offersService.fetch(filters)
       expect(serviceFindSpy).toBeCalled()
     })
 
     test('should return NotifierOfferItem[] on success', async () => {
-      const actualOffers: NotifierOfferItem[] = await offersService.fetch()
+      const actualOffers: NotifierOfferItem[] = await offersService.fetch(filters)
 
       const expectedOffers: NotifierOfferItem[] = MOCK_RESPONSE
         .map(mapFromTransport)
```
vjgene, line 99: can we use a better name than mapFromTransport? this seems to be used in multiple places.

src/api/rif-marketplace-cache/notifier/offers/api.ts
```diff
@@ -56,13 +56,13 @@ class OffersService extends AbstractAPIService
...
-        ...filters,
-        currency: Array.from(filters.currency),
-        channels: Array.from(filters.channels),
-      },
-    })
+    const { currency, channels } = filters
+    const query = filters && {
+      ...filters,
+      currency: currency ? Array.from(currency) : undefined,
+      channels: channels ? Array.from(channels) : undefined,
```
vjgene: why is undefined being used. if it's an array can we not use []?

vjgene (changes requested):
Looks good overall. I just think it'd be more clear from provider perspective  to add an extra message when provider tries to register using same url again for clarity.

itofarina:
@vjgene thanks for reviewing. This PR is still a work in progress as it's a draft 🙂

src/api/rif-marketplace-cache/notifier/offers/api.ts
itofarina: I think we should be able to only specify the needed filters. Like it's better to not specify the currency if you don't want to filter by currency. Does it make sense?

src/api/rif-marketplace-cache/notifier/offers/__tests__/api.test.ts
itofarina: I think this is not related to this PR. Personally, I like this name like it's a standard to parse the DTOs. We can discuss about changing it of course but don't think this PR is the best place to do it 😄.

src/components/pages/notifier/sell/PageWrapper.tsx
itofarina: Actually, if the provider already has an offer we don't show the form, so they won't be able to enter a url.

src/components/organisms/storage/sell/NoMultipleOffersCard.tsx
```diff
@@ -15,13 +15,22 @@ const useStyles = makeStyles(() => ({
...
+const MY_OFFERS_ROUTE: Record<Service, string> = {
+  storage: ROUTES.STORAGE.MYOFFERS.BASE,
+  notifier: ROUTES.NOTIFIER.MYOFFERS.BASE,
+}
+
+type Props = {
+  service: Service
+}
+
+const NoMultipleOffersCard: FC<Props> = ({ service }) => {
```
jurajpiar, line 29: Take it out of the storage folder if it is meant to be reused outside of storage.
Also, it seems rather hacky. Maybe would be better to take path as argument pass route to it directly?

src/components/pages/notifier/sell/NotifierSellPage.tsx
```diff
@@ -29,6 +33,25 @@ const NotifierSellPage: FC = () => {
...
+    const getOwnOffers = async (): Promise<void> => {
+      setIsLoadingOffer(true)
+
+      const offersService = apis['notifier/v0/offers'] as NotifierOffersService
+      offersService.connect(reportError)
+
+      const currentOwnOffers = await offersService.fetch({
+        provider: account,
+      })
+      setHasOffer(Boolean(currentOwnOffers[0]))
```
jurajpiar: Might be safer for runtime to do:
```suggestion
      setHasOffer(Boolean(currentOwnOffers?.length))
```

src/models/marketItems/NotifierFilters.ts
```diff
@@ -10,8 +10,8 @@ type PriceFilter = MinMaxFilter & {
 export type NotifierOffersFilters = {
     size: MinMaxFilter
     price: PriceFilter
-    currency: Set<SupportedTokenSymbol>
-    channels: Set<string>
+    currency?: Set<SupportedTokenSymbol>
+    channels?: Set<string>
```
jurajpiar, line 14: Let me fix the type of the api's _fetch sig, but I think we should keep these enforced as they serve the purpose (probably) and instead type cast upon use.

jurajpiar (changes requested):

jurajpiar (approved):

src/components/organisms/storage/sell/NoMultipleOffersCard.tsx
itofarina: Moved, thanks

src/components/pages/notifier/sell/NotifierSellPage.tsx
```diff
@@ -29,6 +34,25 @@ const NotifierSellPage: FC = () => {
...
+  useEffect(() => {
+    const getOwnOffers = async (): Promise<void> => {
+      setIsLoadingOffer(true)
+
+      const offersService = apis['notifier/v0/offers'] as NotifierCacheAPIService
+      offersService.connect(reportError)
+
+      const currentOwnOffers = await offersService.fetch({
+        provider: account,
+      } as NotifierOffersFilters)
```
vjgene, line 49: What happens if it is the same url but a different provider account? should we also block registration if the url is already registered?

vjgene (approved):
Everything looks good. Just left a note on registering with same url.

src/components/pages/notifier/sell/NotifierSellPage.tsx
itofarina: Thanks, that's a good point! I might be missing something but why you wouldn't be able to provide the same service with another account?
At least in the SC, the only check is against the account ([reference](https://github.com/rsksmart/rif-marketplace-notifier/blob/master/contracts/NotifierManager.sol#L153))
May be @artem-brazhnikov could help us define this behavior?

itofarina: Resolving as it will be tackled on a separated story

PR #876, opened 2021-06-11
fix: fixes styles

itofarina:
Styles are broken in some resolutions, this PR fixes it

### Before
Screen Shot 2021-06-11 at 09 44 56

Screen Shot 2021-06-11 at 10 29 50

### Now
Screen Shot 2021-06-11 at 10 20 28
Screen Shot 2021-06-11 at 10 20 20

### DIFF
--- a/src/components/pages/UncaughtError.tsx
+++ b/src/components/pages/UncaughtError.tsx
@@ -3,22 +3,39 @@ import React, { FC } from 'react'
 import lostAstronaut from 'assets/images/lostAstronaut.png'
 import GridRow from 'components/atoms/GridRow'
 import RoundBtn from 'components/atoms/RoundBtn'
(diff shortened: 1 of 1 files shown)


jurajpiar (approved):
I wonder whether we shouldn't use this PR for the fix of the navigation bar on the error page.
https://rsklabs.atlassian.net/browse/RMKT-611

itofarina:
@jurajpiar Great catch! I was looking into this and would like to tackle it in a separated PR.
We can discuss it but basically, we need to wrap our `ErrorBoundary` component with `withRouter` to listen to route changes and update its status when that happens. To do that we also need to remove this component from the places where there's no router (i.e. `App.tsx`) so it's a bit tricky.

PR #878, opened 2021-06-14
feat(notifier): checks account balance before enabling payment

itofarina:
## Features
- Wraps checkout page with login card and redirects to listing when no order was selected (for example when coming from copy paste url)
- Checks account balance before enabling payment

## Known issues
- Would be best to automatically redirect to listing without asking for the account when there is no offer selected. This improvement could be applied to the logic and flow of storage checkout as well.

## Demo

https://user-images.githubusercontent.com/8449567/121942606-0e2a3d80-cd27-11eb-8b44-5495c4c22330.mov

### DIFF
--- a/src/components/organisms/notifier/buy/CheckoutPayment.tsx
+++ b/src/components/organisms/notifier/buy/CheckoutPayment.tsx
@@ -1,5 +1,7 @@
 import RoundBtn from 'components/atoms/RoundBtn'
-import React, { FC } from 'react'
+import React, {
+  FC, useState, useEffect, useContext,
+} from 'react'
 import Typography from '@material-ui/core/Typography'
 import GridRow from 'components/atoms/GridRow'
 import Grid, { GridProps } from '@material-ui/core/Grid'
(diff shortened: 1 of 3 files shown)


src/components/organisms/notifier/buy/CheckoutPayment.tsx
```diff
@@ -34,6 +40,26 @@ const CheckoutPayment: FC<Props> = ({
...
+  } = tokenXR
+
+  useEffect(() => {
+    const calculateBalance = async (): Promise<void> => {
+      setIsLoadingBalance(true)
+      const balance = await getBalance(
+        web3 as Web3, account as string, selectedTokenSymbol,
+      )
+      setHasEnoughFunds(Big(balance).gte(cryptoPrice))
+      setIsLoadingBalance(false)
```
vjgene: can we catch exception and setIsLoadingBalance(false) in finally?

src/components/pages/notifier/buy/NotifierOfferCheckoutPage.tsx
```diff
@@ -143,7 +144,10 @@ const NotifierOfferCheckoutPage: FC = () => {
   const [txOperationDone, setTxOperationDone] = useState(false)
   const [eventsAdded, setEventsAdded] = useState<NotifierEventItem[]>([])
 
-  if (!order?.item) return null
+  if (!order?.item) {
+    history.push(ROUTES.NOTIFIER.BUY.BASE)
+    return null
```
vjgene, line 149: Is return null required after history.push?

vjgene (changes requested):
Looks great except for couple of things

src/components/pages/notifier/buy/NotifierOfferCheckoutPage.tsx
itofarina: Yes, otherwise it will keep going with the render as `history.push` is an asynchronous call.

src/components/organisms/notifier/buy/CheckoutPayment.tsx
itofarina: Done, thanks!

jurajpiar (approved):
looks good

vjgene (approved):

PR #881, opened 2021-06-16
feat(notifier): adds renew subscription

itofarina:
## Features
- adds renew subscription feature. Enabled only on expired or completed subscriptions
  - interacts with notifier service to renew and get the hash
  - once we have the has, interacts with the SC to renew
- fixes spinner while loading purchased subscriptions. Before, when there were no subscriptions it was showing "No purchases yet" and then the spinner. With this change it shows the spinner until we know if there are pending subscriptions or not.

## Demo
Note: the appearance of the new subscription takes a bit longer because notifier service requires more confirmed blocks than cache.

https://user-images.githubusercontent.com/8449567/124195328-9b4bf100-daa0-11eb-9b2e-437bfe92ab5e.mov

### DIFF
--- a/src/api/rif-notifier-service/models/subscriptions.ts
+++ b/src/api/rif-notifier-service/models/subscriptions.ts
@@ -45,7 +45,7 @@ export type NotificationPreference = {
 export type TopicDTO = {
   notificationPreferences: Array<NotificationPreference>
   type: TopicType
-  topicParams: Array<TopicParams>
+  topicParams?: Array<TopicParams>
 }
 
 export const SUBSCRIPTION_STATUSES = {
--- a/src/api/rif-notifier-service/models/subscriptions.ts
+++ b/src/api/rif-notifier-service/models/subscriptions.ts
@@ -45,7 +45,7 @@ export type NotificationPreference = {
 export type TopicDTO = {
   notificationPreferences: Array<NotificationPreference>
   type: TopicType
-  topicParams?: Array<TopicParams>
+  topicParams: Array<TopicParams>
 }
 
 export const SUBSCRIPTION_STATUSES = {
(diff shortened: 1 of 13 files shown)


src/api/rif-notifier-service/models/subscriptions.ts
```diff
@@ -45,7 +45,7 @@ export type NotificationPreference = {
 export type TopicDTO = {
   notificationPreferences: Array<NotificationPreference>
   type: TopicType
-  topicParams?: Array<TopicParams>
+  topicParams: Array<TopicParams>
```
vjgene: why can't we keep the topicParams optional as before, as they don't apply for NEW_BLOCK, and will also not apply for NEW_TRANSACTIONS when we support it.

src/components/organisms/notifier/details/utils.tsx
```diff
@@ -1,21 +1,22 @@
...
+  topicParams: Array<TopicParams>,
   paramType: TopicParamType,
 ): string => topicParams.find(
   ({ type }) => type === paramType,
 )?.value || ''
 
 export const eventDisplayItemIterator = ({
   notificationPreferences, topicParams, type: eventType,
-}: SubscriptionEvent): SubscriptionEventsDisplayItem => {
+}: TopicDTO): SubscriptionEventsDisplayItem => {
```
vjgene, line 19: To be consistent with other pages, may be we should not use a DTO for display item. what do you think? may be we can have this refactoring in another story

vjgene (approved):
Nice work! Just couple of questions

src/components/organisms/notifier/details/utils.tsx
itofarina: This is a good suggestion, I tried this but the problem is that we need to convert everything back and forth: from notifier response to display item and then from display item to dtos to interact with notifier service again and so on when we want to display latest data...
For now, I would suggest to keep this PR simpler and use only one type that is sufficient for ui display and notifier service interaction and we can tackle this refactor later in another PR. What do you think?

PR #883, opened 2021-06-17
fix(price): fixes precision evaluation for display

jurajpiar:
Resolves https://rsklabs.atlassian.net/browse/RMKT-774

https://user-images.githubusercontent.com/36888576/122475001-ea335a00-cfbb-11eb-9982-53730e6c1f4b.mov

P.S.: the weird price change is due to a bug in cache unrelated to this task


src/utils/priceUtils.ts
```diff
@@ -0,0 +1,36 @@
...
+  precisionType: keyof typeof PRICE_PRECISIONS,
+): string => {
+  const precision = PRICE_PRECISIONS[precisionType]
+
+  const bigPrice = Big(price)
+  const exponent = bigPrice.e
+
+  const fixTo = bigPrice.lt(1) && exponent > precision ? exponent : precision
+
+  return squashTrailingZeroes(bigPrice.toFixed(fixTo))
```
vjgene, line 22: Is it that it appends a zero at the end for numbers like 1.2 is converted to 1.20? is it necessary?

vjgene (approved):
Looks good overall.

src/utils/priceUtils.ts
jurajpiar: No, it doesn't do that. It just squashes trailing zeros to just one zero, as per function name. The residual zero indicates there are no more digits, however, which may be an important distinction as it indicates that the last digit has not been rounded.

PR #885, opened 2021-06-21
feat(notifier): shows only purchased subscriptions

vjgene:
fixes https://rsklabs.atlassian.net/jira/software/projects/RMKT/boards/60/?selectedIssue=RMKT-771

### DIFF
--- a/src/components/pages/notifier/mypurchase/NotifierMyPurchasePage.tsx
+++ b/src/components/pages/notifier/mypurchase/NotifierMyPurchasePage.tsx
@@ -81,6 +81,10 @@ const NotifierMyPurchasePage: FC = () => {
   ).length
   const isAwaitingConfs = Boolean(numberOfConfs)
 
+  const setPurchasedSubscriptions = (allSubscriptions: Array<NotifierSubscriptionItem>): void => {
+    setSubscriptions(allSubscriptions.filter((subscription) => subscription.status !== 'PENDING'))
+  }
+
   useEffect(() => {
     if (account && subscriptionsApi) {
       setIsTableLoading(true)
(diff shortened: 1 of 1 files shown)


src/components/pages/notifier/mypurchase/NotifierMyPurchasePage.tsx
```diff
@@ -81,14 +81,18 @@ const NotifierMyPurchasePage: FC = () => {
...
+
   useEffect(() => {
     if (account && subscriptionsApi) {
       setIsTableLoading(true)
       subscriptionsApi.connect(reportError)
       subscriptionsApi.fetch({
         consumer: account,
       })
-        .then(setSubscriptions)
+        .then(setPurchasedSubscriptions)
```
itofarina: I find it a bit confusing to call it `setPurchasedSubscriptions` only to call `setSubscriptions` and show `subscriptions` prop.
May be it would be cleaner to call it `onSubscriptionsFetch` or something like that?

Also I wonder if this PR would be a good opportunity to get them updated in real time when there is new subscriptions, like the following

```
attachEvent('updated', (updatedValue) => {
  onSubscriptionsFetch(updatedValue)
})
```

jurajpiar: > attachEvent('updated', (updatedValue) => {
  onSubscriptionsFetch(updatedValue)
})

I think this should be done across the service in one PR @farinalvaro. We could create a ticket for it?

jurajpiar (approved):
Please squash

itofarina (approved):
Looks good

PR #892, opened 2021-06-30
feat(notifier): adds email channel

vjgene:
Fixes https://rsklabs.atlassian.net/browse/RMKT-789

### DIFF
--- a/src/utils/stringUtils.ts
+++ b/src/utils/stringUtils.ts
@@ -25,3 +25,5 @@ export const shortChecksumAddress = (address: string): string => shortenString(
 export const removeURLProtocol = (url: string) => url.replace(/^http[s]*:\/\//, '')
 
 export const trailingSlashRegex = new RegExp(/\/$/, '')
+
+export const emailRegex = new RegExp(/^\S+@\S+\.\S+$/)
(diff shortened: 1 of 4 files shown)


vjgene:
https://user-images.githubusercontent.com/11134695/123999455-d384fe00-d9d2-11eb-99c0-46e65208a632.mp4

itofarina (approved):
👌

src/utils/stringUtils.ts
```diff
@@ -25,3 +25,5 @@ export const shortChecksumAddress = (address: string): string => shortenString(
 export const removeURLProtocol = (url: string) => url.replace(/^http[s]*:\/\//, '')
 
 export const trailingSlashRegex = new RegExp(/\/$/, '')
+
+export const emailRegex = new RegExp(/^\S+@\S+\.\S+$/)
```
0xartem, line 29: Alvaro said yesterday that we have this function built-in already, so I came to the conclusion that you shouldn't write your own regex, no?

0xartem (approved):
overall looks good, have just one question

src/utils/stringUtils.ts
vjgene: @artem-brazhnikov as we wanted to keep our check simple i used a regex. If the built in doesn't add lot of complexity like adding external modules, we can use it. @farinalvaro @jurajpiar any ideas where the built in function is?

itofarina: Hey, sorry, I was wrong yesterday. The way it's solved in the library examples is with a regex also.
https://react-hook-form.com/

jurajpiar (approved):

PR #895, opened 2021-07-02
fix(notifier): reuses pending renewal subscription

vjgene:
resolves https://rsklabs.atlassian.net/jira/software/projects/RMKT/boards/60/?selectedIssue=RMKT-794

https://user-images.githubusercontent.com/11134695/124305103-42ed2000-db82-11eb-9310-80e94c1051fc.mov

### DIFF
--- a/src/api/rif-notifier-service/subscriptionUtils.ts
+++ b/src/api/rif-notifier-service/subscriptionUtils.ts
@@ -10,6 +10,7 @@ import { NotifierChannel } from 'models/marketItems/NotifierItem'
 import { SupportedTokenSymbol } from 'models/Token'
 import { UIError } from 'models/UIMessage'
 import { convertToWeiString } from 'utils/parsers'
+import RenewSubscriptionService from './renewSubscription'
 
 const buildBlockEvent = (
   notificationPreferences: Array<NotificationPreference>,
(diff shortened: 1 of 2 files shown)


src/api/rif-notifier-service/subscriptionUtils.ts
```diff
@@ -140,9 +141,26 @@ export const createPendingSubscription = (
...
+    item, [], account,
+  )
+  const renewSubscriptionService = new RenewSubscriptionService(providerUrl)
+  renewSubscriptionService.connect(reportError)
+  return renewSubscriptionService.renewSubscription(
+    subscribeToPlanDTO, subscriptionHash,
+  )
+}
+
+const getOrCreateNewOrRenewalSubscription = async (item: SubscriptionOrderItem,
```
itofarina, line 160: I think this function should be a bit more specific.
As it takes one path or another based on `subscriptionHash` parameter, may be we can we split it into two different functions? Or another option... extract the logic to the parent functions?

src/api/rif-notifier-service/subscriptionUtils.ts
```diff
@@ -152,8 +170,18 @@ export const getOrCreateSubscription = async (item: SubscriptionOrderItem,
...
 }
+
+export const getOrCreateSubscription = (item: SubscriptionOrderItem,
+  eventsAdded: Array<NotifierEventItem>, account: string,
+  reportError: (e: UIError) => void): Promise<SubscriptionSummary> => getOrCreateNewOrRenewalSubscription(item, account, reportError, eventsAdded)
+
+export const getOrCreateRenewalSubscription = (
+  subscriptionHash: string, item: SubscriptionOrderItem, account: string,
+  reportError: (e: UIError) => void,
+): Promise<SubscriptionSummary> => getOrCreateNewOrRenewalSubscription(item, account, reportError, [], subscriptionHash)
```
itofarina, line 187: If they only call another function I think it would be simpler to put the logic here.

itofarina (approved):
Looks good, just think it could be done a bit simpler.

src/api/rif-notifier-service/subscriptionUtils.ts
vjgene: the idea is avoid code duplication. the function getOrCreateNewOrRenewalSubscription contains common code.

src/api/rif-notifier-service/subscriptionUtils.ts
vjgene: This will again result in code duplication. the function getOrCreateNewOrRenewalSubscription contains common code, and not exported. The above are exported functions with more meaningful name.

PR #896, opened 2021-07-05
feat(notifier): shows purchases status

itofarina:
## Features
- Shows subscription status
- Adds tooltip on renew button when disabled

## Screenshot
Screen Shot 2021-07-05 at 17 02 29

### DIFF
--- a/src/components/pages/notifier/mypurchase/mapMyPurchases.tsx
+++ b/src/components/pages/notifier/mypurchase/mapMyPurchases.tsx
@@ -1,6 +1,5 @@
 import React from 'react'
 import { CombinedPriceCell } from 'components/molecules'
-import MarketplaceActionsCell from 'components/molecules/MarketplaceActionsCell'
 import NotificationsBalance from 'components/molecules/notifier/NotificationsBalance'
 import { MySubscription } from 'components/organisms/notifier/mypurchase/PurchasesTable'
 import { NotifierSubscriptionItem } from 'models/marketItems/NotifierItem'
@@ -9,9 +8,13 @@ import MarketplaceAddressCell from 'components/molecules/MarketplaceAddressCell'
 import ExpirationDate, { SubscriptionExpirationType } from 'components/molecules/ExpirationDate'
 import { PlanDTO, PLAN_STATUS } from 'api/rif-marketplace-cache/notifier/offers/models'
 import { SUBSCRIPTION_STATUSES } from 'api/rif-notifier-service/models/subscriptions'
-import { Typography } from '@material-ui/core'
+import Grid from '@material-ui/core/Grid'
+import Typography from '@material-ui/core/Typography'
+import Tooltip from '@material-ui/core/Tooltip'
+import RoundBtn from 'components/atoms/RoundBtn'
 
 const EXPIRATION_WARNING_TRIGGER = 5
+const CANT_RENEW_MESSAGE = 'Subscriptions renewal is available only for completed or expired purchases'
 
 const getExpirationType = (
   { planStatus, daysLeft }: PlanDTO,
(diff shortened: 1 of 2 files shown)


0xartem (approved):
lgtm, thanks @farinalvaro

src/components/pages/notifier/mypurchase/mapMyPurchases.tsx
```diff
@@ -9,8 +8,13 @@ import MarketplaceAddressCell from 'components/molecules/MarketplaceAddressCell'
 import ExpirationDate, { SubscriptionExpirationType } from 'components/molecules/ExpirationDate'
 import { PlanDTO, PLAN_STATUS } from 'api/rif-marketplace-cache/notifier/offers/models'
 import { SUBSCRIPTION_STATUSES } from 'api/rif-notifier-service/models/subscriptions'
+import Grid from '@material-ui/core/Grid'
+import Typography from '@material-ui/core/Typography'
+import Tooltip from '@material-ui/core/Tooltip'
+import RoundBtn from 'components/atoms/RoundBtn'
 
 const EXPIRATION_WARNING_TRIGGER = 5
+const CANT_RENEW_MESSAGE = 'Subscriptions renewal is available only for completed or expired purchases'
```
vjgene, line 17: Can we call it RENEWAL_ERROR_MESSAGE or RENEWAL_BLOCKED_MESSAGE?

Also change the message like

RENEWAL_ERROR_MESSAGE = "Active subscriptions cannot be renewed"

src/components/pages/notifier/mypurchase/mapMyPurchases.tsx
```diff
@@ -48,6 +52,25 @@ const mapMyPurchases = <V extends Function, R extends Function>(
...
+    const renewAction = canRenew
+      ? renewButton
+      : (
+        <Tooltip title={CANT_RENEW_MESSAGE}>
+          <span>
+            {renewButton}
+          </span>
+        </Tooltip>
+      )
+
```
vjgene, line 73: Instead of still showing the renew button, can we jus hide the renew button and only show the message? 

Also can this renew check for subscriptionService.getActiveSubscriptionForUser() and also hide the renewal button, in case there is active subscription? Otherwise the user will try to renew, and upon attempting the message will be shown.

vjgene (changes requested):
Looks good. I think it would be better to do an additional check and also change the way we display renew button

src/components/pages/notifier/mypurchase/mapMyPurchases.tsx
itofarina: This message was taken from [the ticket](https://rsklabs.atlassian.net/jira/software/projects/RMKT/boards/60?selectedIssue=RMKT-796).
Also I don't think it's an error, it's a message on top of a disabled button explaining why it's disabled.

src/components/pages/notifier/mypurchase/mapMyPurchases.tsx
itofarina: How do we let the users know why they can't renew if we hide the option? I think a disabled button with a tooltip message is better in terms of UX.
About having an active subscription, we don't need to call the service again, as here we are fetching from notifier service all subscriptions with status !== PENDING.

If I understood correctly, the other suggestion is to disable all renew buttons from a certain provider if there is one active subscription from that provider, right? I think it's a good one.

itofarina: So after discussing with Artem we created a [new issue](https://rsklabs.atlassian.net/browse/RMKT-817) for this one.

vjgene: About the second point you are right, we need to disable renewal if there are any active subscriptions for the user for a given provider.

vjgene (approved):

PR #897, opened 2021-07-06
fix(notifier): uses subscription expiration instead of plan expiration

vjgene:
https://rsklabs.atlassian.net/browse/RMKT-798

https://user-images.githubusercontent.com/11134695/124575825-03f1ef80-de4c-11eb-8717-fcbd1a85707c.mp4

### DIFF
--- a/src/components/molecules/ExpirationDate.tsx
+++ b/src/components/molecules/ExpirationDate.tsx
@@ -15,7 +15,7 @@ const useStyles = makeStyles(({ palette }: Theme) => ({
     color: palette.warning.main,
   },
   blocked: {
-    color: palette.text.disabled,
+    color: palette.error.main,
   },
 }))
 
--- a/src/components/molecules/ExpirationDate.tsx
+++ b/src/components/molecules/ExpirationDate.tsx
@@ -17,12 +17,16 @@ const useStyles = makeStyles(({ palette }: Theme) => ({
   blocked: {
     color: palette.error.main,
   },
+  disabled: {
+    color: palette.text.disabled,
+  },
 }))
 
 export const EXPIRATION_TYPES = {
   normal: 'normal',
   warning: 'warning',
   blocked: 'blocked',
+  disabled: 'disabled',
 } as const
 
 export type SubscriptionExpirationType = keyof typeof EXPIRATION_TYPES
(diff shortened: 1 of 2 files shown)


src/components/molecules/ExpirationDate.tsx
```diff
@@ -15,7 +15,7 @@ const useStyles = makeStyles(({ palette }: Theme) => ({
     color: palette.warning.main,
   },
   blocked: {
-    color: palette.text.disabled,
+    color: palette.error.main,
```
jurajpiar, line 18: I think it has been agreed to ditch the red colouring in favour of disabling it, unless that requirement had changed recently. @artem-brazhnikov could you confirm either way, please?

jurajpiar (dismissed):
let's check with @artem-brazhnikov about the ux

src/components/molecules/ExpirationDate.tsx
0xartem: @jurajpiar it should be disabled and grey if the plan was deleted. and red if expired

src/components/pages/notifier/mypurchase/mapMyPurchases.tsx
```diff
@@ -7,15 +7,17 @@ import { NotifierSubscriptionItem } from 'models/marketItems/NotifierItem'
...
+import { SUBSCRIPTION_STATUS } from 'api/rif-marketplace-cache/notifier/subscriptions/models'
 
 const EXPIRATION_WARNING_TRIGGER = 5
 
 const getExpirationType = (
-  { planStatus, daysLeft }: PlanDTO,
+  { status, expirationDate }: Pick<NotifierSubscriptionItem, 'status' | 'expirationDate'>,
 ): SubscriptionExpirationType => {
-  if (planStatus !== PLAN_STATUS.ACTIVE || daysLeft <= 0) return 'blocked'
+  const daysLeft = Math.ceil(Math.abs(new Date().getTime() - expirationDate?.getTime()) / 86400000)
```
itofarina:
```suggestion
  const daysLeft = Math.ceil(Math.abs(new Date().getTime() - expirationDate?.getTime()) / 60*60*24*1000)
```

Or create a util function `timestampToDays(milliseconds)`

itofarina (approved):

PR #908, opened 2021-07-14
feat(notifier): shows latest own subs only

jurajpiar:
Resolves https://rsklabs.atlassian.net/browse/RMKT-793

by adding a filter to the incoming subscriptions.

## Notes:
- As mentioned in the code, this is not ideal, as we make the browser work on something that could be done by the cache database (upon a query from the UI).
> This should really be done in SQL (cache) but sequelize doesn't support subquery and raw query would be impractical in this case (src: https://github.com/sequelize/sequelize/issues/5354)

## Screens:
**Before:**
Screenshot 2021-07-14 at 09 35 02
**After, when got expired only:**
Screenshot 2021-07-14 at 09 43 58
**After when got expired and renewed:**
Screenshot 2021-07-14 at 09 37 49

### DIFF
--- a/src/components/pages/notifier/mypurchase/NotifierMyPurchasePage.tsx
+++ b/src/components/pages/notifier/mypurchase/NotifierMyPurchasePage.tsx
@@ -102,7 +102,19 @@ const NotifierMyPurchasePage: FC = () => {
           $ne: SUBSCRIPTION_STATUSES.PENDING,
         },
       })
-        .then(setSubscriptions)
+        .then((incomingSubscriptions: Array<NotifierSubscriptionItem>) => {
+          const filtered = incomingSubscriptions.reduce((acc, subscription) => {
+            const { id, plan: { id: planId } } = subscription
+            const successor = incomingSubscriptions.find(
+              ({ previousSubscription }) => id === previousSubscription,
+            )
+
+            acc[planId] = successor ?? subscription
+            return acc
+          }, {})
+
+          setSubscriptions(Object.values(filtered))
+        })
         .catch((error) => reportError(new UIError({
           id: 'service-fetch',
           text: 'Error while fetching subscriptions.',


src/components/pages/notifier/mypurchase/NotifierMyPurchasePage.tsx
```diff
@@ -102,7 +102,17 @@ const NotifierMyPurchasePage: FC = () => {
           $ne: SUBSCRIPTION_STATUSES.PENDING,
         },
       })
-        .then(setSubscriptions)
+        .then((subs: Array<NotifierSubscriptionItem>) => {
+          // This should really be done in SQL (cache) but sequelize doesn't support subquery and raw query would be impractical in this case (src: https://github.com/sequelize/sequelize/issues/5354)
+          const filtered = subs.reduce((acc, cur) => {
+            const sub: NotifierSubscriptionItem = acc[cur.plan.id] ?? cur
+            acc[cur.plan.id] = cur.expirationDate > sub.expirationDate
+              ? cur : sub
```
0xartem: I would re-write it so it's more readable. but it's not crucial for me now as soon as it works correctly.

0xartem (approved):

src/components/pages/notifier/mypurchase/NotifierMyPurchasePage.tsx
jurajpiar: done

itofarina (commented):
LGTM

itofarina (approved):
LGTM

src/components/pages/notifier/mypurchase/NotifierMyPurchasePage.tsx
```diff
@@ -102,7 +102,19 @@ const NotifierMyPurchasePage: FC = () => {
...
       })
-        .then(setSubscriptions)
+        .then((incomingSubscriptions: Array<NotifierSubscriptionItem>) => {
+          const filtered = incomingSubscriptions.reduce((acc, subscription) => {
+            const { id, plan: { id: planId } } = subscription
+            const successor = incomingSubscriptions.find(
+              ({ previousSubscription }) => id === previousSubscription,
+            )
+
+            acc[planId] = successor ?? subscription
```
vjgene, line 112: when we add multiple active subscriptions per plan, grouping by planid will have issues.

vjgene (changes requested):

src/components/pages/notifier/mypurchase/NotifierMyPurchasePage.tsx
```diff
@@ -102,7 +102,19 @@ const NotifierMyPurchasePage: FC = () => {
...
+            return acc
+          }, {})
+
+          setSubscriptions(Object.values(filtered))
```
vjgene, line 116:
```
.then((incomingSubscriptions: Array<NotifierSubscriptionItem>) => {
          const initial: Array<NotifierSubscriptionItem> = []
          const filtered = incomingSubscriptions.reduce((acc, subscription) => {
            const { id } = subscription
            const successor = incomingSubscriptions.find(
              ({ previousSubscription }) => id === previousSubscription,
            )
            const sub: NotifierSubscriptionItem = successor || subscription

            if (acc.indexOf(sub) === -1) acc.push(sub)
            return acc
          }, initial)
          setSubscriptions(filtered)

```

src/components/pages/notifier/mypurchase/NotifierMyPurchasePage.tsx
jurajpiar: Thought, we're not going to implement multiple active subscriptions per plan.
OK, I'll change that.

src/components/pages/notifier/mypurchase/NotifierMyPurchasePage.tsx
jurajpiar: Thanks for the suggestion. Minor comments: 
1. mutating arrays (acc.push)
2. ~acc.indexOf(sub) === -1~ acc.includes(sub)
3. if (acc.indexOf(sub) === -1) acc.push(sub), this will not work for subs renewed more than once, if I'm not mistaken

vjgene: For point 2 it's also an option to use "const initial: Set<NotifierSubscriptionItem> = new Set<NotifierSubscriptionItem>()
then initial.add(sub)
For point 3. you can only renew a subscription, only once. once renewed that subscription will be hidden, since new subscription is shown instead of the original

jurajpiar:
Waiting for decision https://rsklabs.atlassian.net/browse/RMKT-793

jurajpiar:
Closing in favour of a new solution https://github.com/rsksmart/rif-marketplace-ui/pull/914

PR #910, opened 2021-07-15
feat(notifier-buy): shows only available plans for selection

itofarina:
## Description
Compares the available plans in cache with the active plans in the notifier provider instance to check if the plan is available for new subscriptions.

Resolves [RMKT-828](https://rsklabs.atlassian.net/jira/software/projects/RMKT/boards/60?selectedIssue=RMKT-828)

### DIFF
--- a/src/api/rif-notifier-service/subscriptionPlans.ts
+++ b/src/api/rif-notifier-service/subscriptionPlans.ts
@@ -3,7 +3,7 @@ import { NOTIFIER_RESPONSE_STATUSES } from 'api/rif-notifier-service/models/resp
 import { logNotImplemented } from 'utils/utils'
 import { NotifierAPIService } from './interfaces'
 
-export const address = 'getSubscriptionPlans' as const
+export const address = 'getSubscriptionPlans?activePlansOnly=true' as const
 export type Address = typeof address
 
 export default class SubscriptionPlans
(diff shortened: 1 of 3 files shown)


src/api/rif-notifier-service/subscriptionPlans.ts
```diff
@@ -12,13 +12,19 @@ export default class SubscriptionPlans
...
+    if (!isValidResponse) {
+      this.errorReporter({
+        id: 'service-fetch',
+        text: 'Wrong response from notifier provider',
+        error: new Error(`Wrong response from notifier provider ${JSON.stringify(content)}`),
+      })
+    }
 
-    return isValidResponse && anyActive
+    return content.filter(({ planStatus }) => planStatus === 'ACTIVE')
```
vjgene: we can use 

export const address = '/getSubscriptionPlans?activePlansOnly=true' as const

then

return content

src/api/rif-notifier-service/subscriptionPlans.ts
```diff
@@ -12,13 +12,19 @@ export default class SubscriptionPlans
 
   _fetch = (): Promise<SubscriptionPlanResponse> => this.service.find()
 
-  hasActivePlans = async (): Promise<boolean> => {
+  getActivePlans = async (): Promise<Array<SubscriptionPlanDTO>> => {
```
vjgene, line 15: Considering we are checking the length of the plans to determine 'hasActivePlans' in both ProviderRegistrar and NotifierOffersPage, why can't we return a boolean as before?

vjgene (commented):
Looks good. Just couple of suggestions

src/api/rif-notifier-service/subscriptionPlans.ts
itofarina: Looks much better, will adapt, thanks.

src/api/rif-notifier-service/subscriptionPlans.ts
itofarina: For now we are only using the length, but I just changed this with the idea of comparing the plans we are allowing to select in the checkout with this result.
If the plans are not in this result then they shouldn't be selectable. So we need them for the comparison.

vjgene (approved):
Looks good!

PR #932, opened 2021-07-27
feat(notifier): disables unsupported feature - cancel plan

itofarina:
Adds one unsupported feature that we missed to deactivate


0xartem (approved):

vjgene (approved):
Good catch!

PR #974, opened 2021-09-06
fix(storage): fixes visual defects + buggy mapping

jurajpiar:
Resolves https://rsklabs.atlassian.net/browse/RMKT-914
- fixes the ipfs path in the same view:
Screenshot 2021-09-06 at 14 54 34
 - refactors the renaming of RifAddress to AddressItem back  to original
 - fixes the mapping of the storage agreements, when crypto state is not yet populated, by filtering the agreements before mapping, for example upon reload:
 
https://user-images.githubusercontent.com/36888576/132228509-0b45605b-b948-49d3-a756-4aeee99c9e21.mov

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "rif-marketplace-ui",
-  "version": "1.4.2",
+  "version": "1.4.3",
   "lockfileVersion": 1,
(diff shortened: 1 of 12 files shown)


0xartem (approved):
lgtm

PR #977, opened 2021-09-09
fix(notifier-buy): shows only active plans

itofarina:
Shows only the plan options that are present in the notifier instance.


jurajpiar (approved):
LG

